PR #2, opened 2017-09-13
Refactor and Divide some function

Ikuyadeu:
17592bd is just rename and add new line for lintr.
26dac1b is divide some function by messages kind(not important and reset is ok).


R/document.R
```diff
@@ -0,0 +1,107 @@
+workspace <- R6::R6Class("textDocument",
```
randy3k: why do we need these refactoring? Are you planning to move all the handlers here?

R/server.R
```diff
@@ -2,11 +2,11 @@
 run <- function(debug = FALSE, stdin = "stdin", stdout = "stdout") {
     logger$set_mode(debug = debug)
     langserver <- LanguageServer$new(stdout = stdout)
-    con <- file(stdin)
-    open(con, blocking = TRUE)
+    content <- file(stdin)
```
randy3k: thank for the cleaning here.

R/workspace.R
```diff
@@ -0,0 +1,21 @@
+workspace <- R6::R6Class("workspace",
```
randy3k: again, why do you think a R6 class is necessary?

R/document.R
Ikuyadeu: Right, I added class for identify `completionItem` and `textDocument`.
But dividing by file may be better than this?

R/workspace.R
Ikuyadeu: Adding this class is for consistency `document.R` file, but don't need.

R/document.R
randy3k: That's what I think. A R6 class does not seem necessary.

Ikuyadeu: Thanks, but I can't fix this now soon.
I can fix this here's morning(9~ hours ago)

randy3k: No problem.

PS: I think you meant 9 hr later.

Ikuyadeu:
@randy3k Ok, fixed.

R/codeLens.R
```diff
@@ -0,0 +1,3 @@
+resolve  <- function(self, id, params) {
```
randy3k: we cannot just call it `resolve`, still need a prefix.

R/documentLink.R
```diff
@@ -0,0 +1,3 @@
+resolve  <- function(self, id, params) {
```
randy3k: same as another resolve

R/completionItem.R
```diff
@@ -0,0 +1,3 @@
+resolve  <- function(self, id, params) {
```
randy3k: same as another resolve

R/document.R
```diff
@@ -0,0 +1,81 @@
+did_open <- function(self, params) {
```
randy3k: may be still a good idea to have the prefix document? like `document_did_open`?

R/document.R
```diff
@@ -0,0 +1,81 @@
...
+    textDocument <- params$textDocument
+    publish_diagnostics(self, textDocument$uri)
+}
+
+did_change <- function(self, params) {
```
randy3k: I saw you added comments for identify the functions as notification or request, may be we should also do the same thing here?

Ikuyadeu:
Thank you, fixed.

R/document.R
Ikuyadeu: I think not so, because we can identify from file name.

randy3k: But it is not how a R package works, all the functions are loaded to the same namespace no matter how the files are named. For instance, if there is another `did_open` in `workspace.R`. One of them will be overwritten by the other.

randy3k:
You also need to change the name of the functions [here](https://github.com/REditorSupport/languageserver/blob/master/R/zzz.R#L9) and remove `notification_handlers` and `request_handler`.

R/document.R
Ikuyadeu: I see, so I'll fix to `text_document_did_open`. (from https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md)
Is it OK?

R/general.R
```diff
@@ -0,0 +1,31 @@
...
+    # TODO: result lint result of the package
+    # lint_result <- lintr::lint_package(rootPath)
+}
+
+# Request
+on_shutdown <- function(self, id, params) {
+
+}
+
+# Request
```
randy3k: it is a notification.

R/general.R
```diff
@@ -0,0 +1,31 @@
...
+on_exit <- function(self, params) {
+    self$will_exit <- TRUE
+}
+
+# Request
+cancel_request <- function(self, params) {
```
randy3k: it is also a notificaiton

R/general.R
```diff
@@ -0,0 +1,31 @@
+on_initialize <- function(self, id, params) {
```
randy3k: `general.R` may be indeed too general. How about moving them to sub folders `handlers`?

Ikuyadeu: You mean this https://github.com/Ikuyadeu/languageserver/tree/refactor/R/handlers?

Ikuyadeu:
more fix

Ikuyadeu:
> general.R may be indeed too general. How about moving them to sub folders handlers?
 
You mean this https://github.com/Ikuyadeu/languageserver/tree/refactor/R/handlers?

Ikuyadeu:
How is this folder

handlers/
  |--request/
  |              |--document.R
  |              |--workspace.R
  |     
  |--notification/
  |              |--document.R
  |              |--workspace.R

R/zzz.R
```diff
@@ -1,11 +1,10 @@
-request_handlers <- list(
```
randy3k: You should not change `request_handlers` and `notification_handlers`, they are being dispatched at
https://github.com/REditorSupport/languageserver/blob/44050bbefe376ee66b2625a61e7954c49f5e1098/R/languageserver.R#L49-L72

randy3k:
I think the current structure (`handlers/document.R`, `handlers/workspace.R`)... is fine

Ikuyadeu:
OK, how is it?

randy3k:
Sorry @Ikuyadeu 
I gave you wrong information, the R files have to be in the top level of the `R` directory.
Like this https://github.com/REditorSupport/languageserver/tree/master/R

Ikuyadeu:
Oops, I'll be careful next.

randy3k:
No problem. Thank you for the work.

PR #16, opened 2018-05-11
Provide settings needed for Emacs

izahn:
I just tried this with Emacs + ESS, with great success. This PR documents the configuration needed for Emacs integration.


randy3k:
Great. I was waiting for an eMac user like you.

PR #17, opened 2018-05-26
Use more effecient methods to construct completions

renkun-ken:
1. Use `startsWith` in vectorized way and `lapply` for higher efficiency
2. Fix error when typing `for (i` in editor which causes the following error:

```
[Error - 11:12:35 PM] Request textDocument/completion failed.
  Message: non-character object(s)startsWith(names(args), token)

  Code: -32603 
```

### DIFF
--- a/R/completion.R
+++ b/R/completion.R
@@ -24,31 +24,19 @@ CompletionItemKind <- list(
 
 package_completion <- function(token) {
     installed_packages <- rownames(utils::installed.packages())
-    completions <- list()
-
-    for (package in installed_packages) {
-        if (startsWith(package, token)) {
-            completions <- append(completions, list(list(
-                label = package,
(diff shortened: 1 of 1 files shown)


randy3k:
thanks. it seems reasonable. will pull and test it later.

randy3k:
hey @renkun-ken,
are you interested to be a member of REditorSupport?

renkun-ken:
Sure! I'm trying to work with vscode but I'm not familiar with how LSP works and how to correctly debug the package at the moment. Will read the LSP docs and more of the code later.

randy3k:
Welcome to our team. I am glad that there are more contributors. Unfortunately, I personally do not use VSCode and don't know much how to debug it.

I would be happy to assist future contributors on Atom, Sublime Text or Neovim though.

PR #22, opened 2018-06-26
completions from document

randy3k:


R/completion.R
```diff
@@ -105,14 +105,57 @@ completion_reply <- function(id, uri, workspace, document, position) {
 
     completions <- list()
 
+    expr <- attr(document, "expr")
+
     if (nzchar(token)) {
+        if (length(expr)) {
```
randy3k: I guess we could do these parsings in `text_document_did_open` and `text_document_did_save`. It may be redundant to redo all these operations if the context is not changed.

R/handlers-textsync.R
```diff
@@ -11,11 +14,15 @@ text_document_did_open <- function(self, params) {
...
+    text <- contentChanges[[1L]]$text
     uri <- textDocument$uri
-    doc <- stringr::str_split(text, "\n")[[1]]
+    doc <- strsplit(text, "\n", fixed = TRUE)[[1L]]
     # remove last empty line
-    if (nchar(doc[[length(doc)]]) == 0) doc <- doc[-length(doc)]
+    if (!nzchar(doc[[length(doc)]])) {
+        doc <- doc[-length(doc)]
+    }
+    expr <- tryCatch(parse(text = doc, keep.source = FALSE), error = function(e) NULL)
```
randy3k: I am not sure if we want to do these every time the content changes (maybe too much overheads?). Only `text_document_did_open` and `text_document_did_save` may be sufficient?

renkun-ken:
The branch is rebased on latest `master` and the parsing is reduced to happen only on document open and save.

randy3k:
Thanks for the input. I'll try to review it tomorrow.

randy3k:
Hi @renkun-ken,

There is actually an cleaner approach. We could create an anonymous namespace using https://github.com/REditorSupport/languageserver/blob/59aa4fe2ea6ce2395a2c36b941ee8fccc4c3de8b/R/workspace.R#L1

and `eval` the expressions in the namespace. In this way, the completion and signature will automatically go through.

I don't want to discredit your hard work, so I will merge your work but revert them. I hope you will be okay with it.

renkun-ken:
Thanks! I thought evaluation is not desired to provide completion and signature. But I'm not sure how `Namespace` is supposed to work. A cleaner approach is much appreciated. Thanks for your notice!

PR #23, opened 2018-06-27
Require latest styler release for more informative error message

lorenzwalthert:
I was happy to learn about styler being used in this project 😊. In https://github.com/r-lib/styler/issues/396, a user of your project mentioned that styler fails if Windows style EOL characters are used. We have decided to not support Windows style EOL at the moment, but instead ask the user to use Unix style EOL if they want to use styler. Windows style line endings now throw an informative error message with version `v1.0.2`. Hence, I suggest you make this minimal version requirement for the dependency styler.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -17,7 +17,7 @@ Imports:
     R6,
     repr,
     stringr,
-    styler,
+    styler (>= 1.0.2),
     tools
 Suggests:
     testthat


randy3k:
Thanks for the PR. I guess we also should be better support of the windows style newline, i.e., split the text with `\r\n` [rather than `\n`](https://github.com/REditorSupport/languageserver/blob/eb270a754d06140ee76bbdd1646f4b7289b7f555/R/handlers-textsync.R#L16-L16).

lorenzwalthert:
Unfortunately styler is not going to support windows newline characters most likely. But if you supply a vector here each element corresponds to a line of code, styler can also handle it, is it es noch it's have to ne a single string.

PR #31, opened 2019-02-23
fix CRAN checks

Tutuchan:


Tutuchan:
We should setup CI with Travis for example, to prevent this kind of problems in the future. I can take care of it.

PR #37, opened 2019-06-15
add go to definition

### LINKED ISSUE #34: Go to Definition
It seems like there is some interest in Go to Definition lately (e.g., https://github.com/Ikuyadeu/vscode-R/issues/105) so I thought I'd open an Issue for it.

To me it looks like the way you'd go about implementing it would be to have `workspace_sync` grab the function definitions in the documents and store their locations. This would be done either by modifying `parse_document` or creating a similar function. To look up the location of a requested definition, you'd use a function that would be structured like `get_signature`.

Functions with the same name can be declared more than once in R; in these cases, I think provided Go to Definition goes to one of the definitions it would be okay.

Conceivably you could have Go to Definition work for variables as well as functions but since they're more likely to be mutated I don't think it's worth including them at this point.

@randy3k How does the above sound to you? I thought I might have a try at implementing it if no-one else  is working on it. (@minkir014 @Tutuchan If you're working on this already let me know - I'd be happy to leave it to someone else!)

andycraig:
Closes #34

This PR adds Go to Definition functionality.

@Tutuchan added the infrastructure and code to go to definitions of functions in packages; I added code to go to definitions in local files.

The following rules are used:
1. Go to definition on a function with a `pkg::` qualifier will look for that function first in package `pkg` and then in local files;
1. Go to definition on a function without a `pkg::` qualifier will look for that function first in local files and then in loaded packages.

Note that the language server only becomes aware of functions in local files as those files are registered via `didOpen` etc. A future feature that registers, e.g., all files in the current project would make Go to Definition more useful.

@Tutuchan I have rebased your commits and my commit onto `master`.

Running the unit tests using `devtools::test()` requires that the package already be installed (not just with `devtools::load_all()`). Otherwise, the unit tests hang. This seems to be the case with `master` too, though.


minkir014:
Do you know why the last commit failed....

minkir014:
And does this package goes to definitions for all packages that are installed???

andycraig:
Hi @minkir014, the checks failed for two reasons:

1. The documentation didn't quite meet CRAN requirements, and
1. Some unit tests were failing because the language server takes a bit longer to run in Travis CI and I had it requesting data before the data was ready.

Happily, both those issues seem to be resolved now.

For packages, there isn't really a 'definition' to go to. So in the `library(shiny)` example it wouldn't do anything. But, if you clicked on `runExample` in `shiny::runExample` it _would_ go to the definition for `runExample`.

randy3k:
@andycraig excellent work. I will take a look as soon as possible. @Tutuchan feels free too give any comments.

minkir014:
> Note that the language server only becomes aware of functions in local files as those files are registered via `didOpen` etc. A future feature that registers, e.g., all files in the current project would make Go to Definition more useful.

andycraig:
@randy3k @Tutuchan Thanks both! @Tutuchan Your code was very clear and easy to continue on with :)

andycraig:
@minkir014 For functions in local files (not packages), that’s correct. This is the same for other features like Code Completion and Hover.

kforner:
> Yes, I don't know why it behaves that way but I noticed the same behavior.

I happened to me. I think this is because the R process run by https://github.com/REditorSupport/languageserver/blob/master/R/languageclient.R#L18 does not find the package, since it is not installed. But there is not any error check at this point. 
I propose to add a sleep, then a check that the process is still alive, otherwise raise an execption with the error stream content.

andycraig:
Hi @kforner, thanks for looking at this. I created a separate issue for it (#39).

randy3k:
I don’t think I will have time review it in details in the near future. After I quick skim, I guess we could merge it first and evaluate.

andycraig:
@randy3k Sounds good. If you have comments in the future, let me know and I'll be happy to make fixes etc. as required :+1:

PR #43, opened 2019-07-14
fix handling of multiple function removal from files

### LINKED ISSUE #41: Problem when multiple functions are removed from file
There is unfortunately a problem with some of the Go to Definition code I added, which means that the `DefinitionCache` class does not update correctly when multiple functions are removed from a file at the same time. This means that Go to Definition can send the user to a location in a file where a function previously existed but now does not. In some cases I think it can crash the Language Server with an error like:

    Error in `[[<-`(`*tmp*`, removed_functs, value = NULL) : 
      no such index at level 1

I haven't been able to reproduce this error reliably though.

It's possible this is solely a result of using `[[]]` rather than `[]` at https://github.com/REditorSupport/languageserver/blob/master/R/definition.R#L18

I will have some free time to look into this properly in a couple of weeks.

andycraig:
Fixes #41 

I added a unit test to `test-definition.R` that simulates two functions being simultaneously removed from a file. It fails when run with the pre-PR version of `definition.R`, as expected based on #41. It passes when run with the PR version of `definition.R`.

### DIFF
--- a/R/definition.R
+++ b/R/definition.R
@@ -15,7 +15,7 @@ DefinitionCache <- R6::R6Class("DefinitionCache",
             functs <- names(ranges)
             removed_functs <- setdiff(private$uris[[uri]], functs)
             if(!is.null(removed_functs) && length(removed_functs) > 0) {
-                private$definitions[[removed_functs]] <- NULL
+                private$definitions[removed_functs] <- NULL
             }
             for(funct in functs) {
                 private$definitions[[funct]] <- location(uri, ranges[[funct]])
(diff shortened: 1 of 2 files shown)


randy3k:
Thanks, I am going to merge this now and review the definition feature as a whole later.

andycraig:
Sure thing @randy3k, thank you!

randy3k:
@andycraig 

I haven't tested the definition feature very thoroughly, but I am very happy with the current results that we have. (Thank you for writing the tests too).

Do you think we are ready to make a release now?

andycraig:
@randy3k Great! I've used it quite a lot and haven't noticed any further problems specific to Go to Definition. I have had some problems in VSCode working with files containing non-ASCII characters, but I haven't been able to create a small reprex. If you're happy to leave that to another day, I would say it's ready for release. 

It sounds like the CRAN servers can be slow so to avoid trouble there I've created a PR (#45) that will skip the time-based tests on CRAN.

Issue #55, opened 2019-09-18
Recursive parsing for symbols in global scope


renkun-ken:
I'm trying to implement recursive parsing for symbols in global scope. For example, in the following script:

```r
x0 <- 0

if (x0 > 0) {
  x1 <- 1
  if (x1 < 0) {
    x2 <- 2
  } else {
    x3 <- 3
  }
}

for (i in 1:10) {
  x4 <- 4
  if (i >= 5) {
    x5 <- 5
  }
}

while (i < 100) {
  x6 <- 6
  if (x6 < 10) {
    x7 <- 7
    {
      x8 <- 8
    }
  }
}
```

All expressions are evaluated in the global scope so `x1 .. x8` would be in the same scope if defined.

PR #56, opened 2019-09-18
Recursive parsing

renkun-ken:
This PR is an initial attempt to address #55 by implementing recursive parsing of a document so that symbols in the global scope would be added to the completion list.

As for function definition range, since R's `parse()` only provides document ranges of top-level expressions, so that non-top level functions defined will share the same definition range at the moment.

### DIFF
--- a/R/document.R
+++ b/R/document.R
@@ -255,8 +255,9 @@ resolve_package_dependencies <- function(pkgs) {
             callr::r(
                 function(pkgs) {
                     for (pkg in pkgs) {
-                        library(pkg, character.only = TRUE)
-                    }                
+                        tryCatch(library(pkg, character.only = TRUE), 
+                            error = function(e) NULL)
+                    }
                     search()
                 },
                 list(pkgs = pkgs)
--- a/R/document.R
+++ b/R/document.R
@@ -229,8 +229,8 @@ parse_expr <- function(expr, env, level = 0L, srcref = attr(expr, "srcref")) {
                 # R is 1-indexed, language server is 0-indexed
                 first_line <- cur_srcref[1] - 1
                 first_char <- cur_srcref[5] - 1
-                last_line <- cur_srcref[3]
-                last_char <- cur_srcref[6]
+                last_line <- cur_srcref[3] - 1
+                last_char <- cur_srcref[6] - 1
                 definition_range <- range(
                     position(first_line, first_char),
                     position(last_line, last_char)
--- a/R/document.R
+++ b/R/document.R
@@ -174,6 +174,7 @@ parse_document <- function(path) {
     }
     env <- parse_env()
     parse_expr(expr, env)
+    env$packages <- resolve_package_dependencies(env$packages)
     env
 }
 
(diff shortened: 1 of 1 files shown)


renkun-ken:
I profiled the performance of `parse_expr` with my production code of 1k+ lines and nearly 10 dependent packages and find everything works super fast except for the `callr::r()` part that resolves package dependencies.

Repeatedly calling `callr::r()` starts R session so it is sure to be slow. I change the code so that all packages are resolved only once and the timing drops from 2s+ to 0.7s in my case.

randy3k:
👍 will take some time to look at it tomorrow.

R/document.R
```diff
@@ -172,66 +172,104 @@ parse_document <- function(path) {
...
-         signatures = signatures, formals = formals, definition_ranges = definition_ranges)
+    env
+}
+
+resolve_package_dependencies <- function(pkgs) {
+    if (length(pkgs)) {
+        deps <- tryCatch(
+            callr::r(
+                function(pkgs) {
+                    for (pkg in pkgs) {
```
randy3k: maybe we could a bit smarter to avoid re-examine packages which are already loaded. For example, `pkgs = c("tidyverse", "ggplot2")`. After checking the dependencies of `tidyverse`, there is no need to examine the dependencies of `ggplot2` since `ggplot2` is a dependency of `tidyverse`.

randy3k:
@renkun-ken I haven't tested it. Is it ready to be fully reviewed?

renkun-ken:
@randy3k Yes, please :). I've been using it for production quite a while.

R/document.R
```diff
@@ -172,66 +172,107 @@ parse_document <- function(path) {
...
+                signature <- utils::capture.output(print(signature[1:2]))
+                signature <- paste0(trimws(signature, which = "left"), collapse = "\n")
+                signature <- trimws(gsub("NULL\\s*$", "", signature))
+                env$signatures[[funct]] <- signature
+                # R is 1-indexed, language server is 0-indexed
+                first_line <- cur_srcref[1] - 1
+                first_char <- cur_srcref[5] - 1
+                last_line <- cur_srcref[3] - 1
+                last_char <- cur_srcref[6] - 1
+                definition_range <- range(
```
randy3k: It doesn't seem that `cur_srcref` contains the right information if the doc contains multi bytes chars.

renkun-ken: According to [documentation](https://www.rdocumentation.org/packages/base/versions/3.6.1/topics/srcfile), `srcref` is a vector of the following information:

```
c(first_line, first_byte, last_line, last_byte, first_column,
  last_column, first_parsed, last_parsed)
```

> Bytes (elements 2, 4) and columns (elements 5, 6) may be different due to multibyte characters. If only four values are given, the columns and bytes are assumed to match. Lines (elements 1, 3) and parsed lines (elements 7, 8) may differ if a #line directive is used in code: the former will respect the directive, the latter will just count lines. If only 4 or 6 elements are given, the parsed lines will be assumed to match the lines.

I test it by parsing the following code where multibyte characters present:

```r
函数1 <- function(参数1) {
  参数1 + 1
}
   函数2 <- function(参数1, 参数2) {
  参数1 + 参数2
}
```

And I look at its `srcref` with 

```r
e <- parse("test.R", keep.source=TRUE)
s <- attr(e, "srcref") 
str(s)
```

it looks like 

```
List of 2
 $ : 'srcref' int [1:8] 1 1 3 1 1 1 1 3
  ..- attr(*, "srcfile")=Classes 'srcfilecopy', 'srcfile' <environment: 0x7f8c6a58d958> 
 $ : 'srcref' int [1:8] 4 4 6 1 4 1 4 6
  ..- attr(*, "srcfile")=Classes 'srcfilecopy', 'srcfile' <environment: 0x7f8c6a58d958> 
```

And the result looks correct. In VSCode, the definition ranges are highlighted, and I can jump to each of them. And the ranges and jump position all look right. I think the four numbers are correct as definition ranges in multibyte cases.

randy3k: In R, that number is the position in terms of the number of bytes of the characters in UTF-8.

I just read again the documentation of LSP

> The current protocol is tailored for textual documents whose content can be represented as a string. There is currently no support for binary documents. A position inside a document (see Position definition below) is expressed as a zero-based line and character offset. The offsets are based on a UTF-16 string representation. So a string of the form a𐐀b the character offset of the character a is 0, the character offset of 𐐀 is 1 and the character offset of b is 3 since 𐐀 is represented using two code units in UTF-16. To ensure that both client and server split the string into the same line representation the protocol specifies the following end-of-line sequences: ‘\n’, ‘\r\n’ and ‘\r’.

It seems that it is more complicated than this. The character offset is the number of bytes of a character in UTF-16. (All chinese chars are 2 bytes in UTF-16, we didn't discover the issue because that most, but not all, chinese chars are 2bytes in utf-8)

It is not only an issue for this piece of code, but a more general issue about `position`.

randy3k: Basically, the use of `nchar` is wrong in many cases.

renkun-ken: I'm not an expert on multi-byte issues. Maybe it's worth a new issue to address the multibyte problem? I just find that no multibyte symbols are in auto-completion list. There must be something wrong somewhere.

randy3k: I will merge it first and fix the issues all together.

PR #61, opened 2019-09-20
clean test code

randy3k:


andycraig:
@randy3k Thanks for the test code cleanup!

Issue #67, opened 2019-09-23
Request textDocument/rangeFormatting failed


renkun-ken:
Create a document like the following:

```r
cols <- c("a", "b", "c", "d", "e",
  "f", "g", "h", "i", "j", "k", "l",
  "m", "n", "o", "p")

cols_new <- test("file1",
  names = c(1:10, 15:20),
  colnames = c())
```

And then copy the elements of `cols` (from "a" to "p" without `c()`) and paste them into the parentheses in `colnames = c()` and an internal error of languageserver occurs:

```
internal error:  <text>:3:22: unexpected ')'
2:   "f", "g", "h", "i", "j", "k", "l",
3:   "m", "n", "o", "p"))
                        ^

[Error - 4:14:43 PM] Request textDocument/rangeFormatting failed.
  Message: <text>:3:22: unexpected ')'
2:   "f", "g", "h", "i", "j", "k", "l",
3:   "m", "n", "o", "p"))
                        ^

  Code: -32603 
```

randy3k:
It is VSCode specific issue, but there is more going on here. The problem is, in `range formatting`, we always get the full lines of the selection.
https://github.com/REditorSupport/languageserver/blob/94a8db03369cf393606c4a570689707a68ba7bc4/R/formatting.R#L77-L80

When you paste "a" to "p", the text being formatted is
```
  colnames = c("a", "b", "c", "d", "e",
  "f", "g", "h", "i", "j", "k", "l",
  "m", "n", "o", "p")))
```

That's why we get an error.

The easiest fix is to catch the possible error of `style_text` since we cannot predict what will be passed to it.

Does VSCode really expect to run formatting for every paste? Sometime user may copy and paste an incomplete piece of code and obviously we don't want to format it.

renkun-ken:
In vscode, there are `formatOnType`, `formatOnPaste`, and `formatOnSave` options. I think such formatting error can be suppressed unless debug mode is on.

randy3k:
I believe that the error is suppressed when debug mode is off, isn't it?

renkun-ken:
My debug mode is off but this still appears.

randy3k:
Note that you will need to restart VSCode.

randy3k:
The problem is more complicated when I thought.

```
foo <- function(
		x = 1,
		y = 2
	) {
	# body
}

```
If I paste `z = 1` to replace `y = 2`. I got

```
foo <- function(
		x = 1,
		z <- 1
[...]

PR #71, opened 2019-09-23
Use line_breaks scope in range formatting

renkun-ken:
This PR attempts to address #67 in a more robust way. Range formatting always applies `scope = "line_breaks"` formatting which does everything but not replaces tokens (no replace `<-` with `=`), while document formatting applies `scope = "tokens"` which must be context-aware.

This makes the discussed examples all work well.

### DIFF
--- a/R/formatting.R
+++ b/R/formatting.R
@@ -30,10 +30,12 @@ style_file <- function(path, options) {
 #' @param text a vector of text
 #' @param options a named list of options, with a `tabSize` parameter
 #' @param indentation amount of whitespaces put at the begining of each line
-style_text <- function(text, options, indentation = "") {
+style_text <- function(text, options, scope, indentation = "") {
     new_text <- styler::style_text(
         text,
-        transformers = styler::tidyverse_style(indent_by = options$tabSize)
+        transformers = styler::tidyverse_style(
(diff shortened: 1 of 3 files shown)


randy3k:
I am not sure about it. As a past vim user, I am quite sure that most vim users would format a paragraph instead the whole document. This PR ultimately disable them to format a paragraph and transform and assignment operators.

Though my PR is not better in any way.

renkun-ken:
I guess maybe we can leave the scopes of document formatting and range formatting as options to users to allow the following levels of scopes as described at <https://styler.r-lib.org/reference/tidyverse_style.html#details>:

> * "none": Performs no transformation at all.
> * "spaces": Manipulates spacing between token on the same line.
> * "indention": In addition to "spaces", this option also manipulates the indention level.
> * "line_breaks": In addition to "indention", this option also manipulates line breaks.
> * "tokens": In addition to "line_breaks", this option also manipulates tokens.

Also, some other options of the styler can be offered too.

What do you think?

randy3k:
Perhaps if partial line is selected, we apply line_break scope, otherwise we apply token_scope?

renkun-ken:
I think your suggestion makes sense and works for me.

PR #70, opened 2019-09-23
Make to_string() more robust

renkun-ken:
The elements of `dots` can be anything. This PR makes it more robust.

PR #73, opened 2019-09-24
Fix to_string for handling complex objects

renkun-ken:
Passing objects such as `environment` to `to_string()` causes error since `jsonlite::toJSON` cannot handle it.

This PR fixes this by directly using the printing content of non-atomic, non-list objects.

### DIFF
--- a/R/log.R
+++ b/R/log.R
@@ -23,7 +23,7 @@ to_string <- function(...) {
                         ""
                     }
                 } else {
-                    utils::capture.output(print(x))
+                    paste0(utils::capture.output(print(x)), collapse = "\n")
                 }
             }, character(1L)
         )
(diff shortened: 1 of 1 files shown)


randy3k:
It's getting more sophisticated.

renkun-ken:
As I comment at <https://github.com/REditorSupport/languageserver/pull/70#issuecomment-534399816>, the input object may be `Classes 'Namespace', 'R6' <Namespace>`, an `R6 object` which is essentially an environment. The most original version works because `sapply` simplifies all texts into a character vector in which the environment actually already causes error but captured.

I think it makes sense to only convert those convertible objects to JSON and leave others on their own text representation.

randy3k:
Anyway, we only need this function to work and it is not important how it works. Merge it now.

renkun-ken:
Indeed. Thanks!

PR #74, opened 2019-09-24
fix some strange behavior of formatting on paste

randy3k:


renkun-ken:
I tested this PR and it works very well.

Issue #57, opened 2019-09-18
Provide symbols in local scope in auto-completion


renkun-ken:
For the following cases, I'm wondering if it is possible to provide local symbols in the function body of `test0` and the expression of `test1`.

```r
test0 <- function(x, y, z) {
  x + y + z
}

test1 <- local({
  p <- 1
  q <- 2
  if (q >= p) {
    cat("hello!\n")
  }
})
```

randy3k:
It would be lovely to have. Sure enough, it is not that simple, for example

```r
foo <- function() {
	x <- 1
	bar <- function() {
		y <- 2
			bla <- function() {
				z <- 3
			}
	}
}
```


At some point, I think we could move the parser to a separate package. Maybe even use RCpp to move some code to c++.

randy3k:
codetools may be useful here https://cran.r-project.org/web/packages/codetools/index.html

renkun-ken:
I notice `getParseData()` can be very useful to provide full-text parser info. For example, the following code:

```r
x0 <- 1
x1 <- 1 + x0
x2 <- 2 + x0 + abs(x0)

f <- function(x, y) {
  x + y
}

g <- function(x) {
  y <- 1
  bar <- function(z) {
    x + y + z
  }
}
```

And then we can use `getParseData` to generate a data frame that maps each code unit to its location by reframing `attr(e, "srcfile")$parseData`.

```r
e <- parse("test1.R", keep.source = TRUE)
getParseData(e)
```

```
    line1 col1 line2 col2  id parent                token terminal     text
7       1    1     1    7   7      0                 expr    FALSE         
1       1    1     1    2   1      3               SYMBOL     TRUE       x0
3       1    1     1    2   3      7                 expr    FALSE         
2       1    4     1    5   2      7          LEFT_ASSIGN     TRUE       <-
4       1    7     1    7   4      5            NUM_CONST     TRUE        1
5       1    7     1    7   5      7                 expr    FALSE         
20      2    1     2   12  20      0                 expr    FALSE         
10      2    1     2    2  10     12               SYMBOL     TRUE       x1
12      2    1     2    2  12     20                 expr    FALSE         
11      2    4     2    5  11     20          LEFT_ASSIGN     TRUE       <-
19      2    7     2   12  19     20                 expr    FALSE         
13      2    7     2    7  13     14            NUM_CONST     TRUE        1
14      2    7     2    7  14     19                 expr    FALSE         
[...]

PR #82, opened 2019-10-07
Initial implementation of scope completion using xpath

renkun-ken:
This is an initial attempt to address #57 using `xmlparsedata` and XPath selectors.

To provide completions in each local scope, all enclosing scopes are detected (except for the top-level). In each of them, all symbols in the assignment expression (`<-`, `->`, and `=`) and function formals in each of these enclosing scopes are collected as locally available symbols at the current position.

The following images demonstrates the scope completions:

image

image

image

One thing yet to improve is that adding new lines to document or removing lines from document make parsed document out-of-sync, so that the line positions are out-of-sync. We should consider parse documents more frequently (i.e. trigger by adding/removing lines) without loading packages (which is the main overhead. parsing is very fast in most cases).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,7 +22,9 @@ Imports:
     stringr,
     styler (>= 1.0.2),
     tools,
-    utils
+    utils,
+    xml2,
+    xmlparsedata
 Suggests:
     processx,
     knitr,
--- a/R/completion.R
+++ b/R/completion.R
@@ -102,6 +102,42 @@ workspace_completion <- function(workspace, token, package = NULL, exported_only
     completions
 }
 
+find_enclosing_scopes <- function(x, line, col) {
+    xpath <- sprintf("//expr[(@line1 < %d or (@line1 = %d and @col1 <= %d)) and 
+        (@line2 > %d or (@line2 = %d and @col2 >= %d))]",
+        line, line, col, line, line, col)
+    xml2::xml_find_all(x, xpath)
+}
+
+scope_completion <- function(uri, workspace, token, position) {
+    xml_doc <- workspace$get_parse_data(uri)$xml_doc
+    if (is.null(xml_doc)) {
+        return(list())
+    }
+
+    enclosing_scopes <- find_enclosing_scopes(xml_doc, 
+        position$line + 1, position$character + 1)
+    
+    symbol_formals <- xml2::xml_text(xml2::xml_find_all(enclosing_scopes, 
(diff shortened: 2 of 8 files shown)


randy3k:
Awesome. Thanks for working on it. I think eventually we could move the code of detecting function definitions to the same xml approach. Do you think we could predetermine some useful information in `parse_document`?

renkun-ken:
@randy3k Indeed. I believe we could eventually move all parse-related functionalities toward this xml approach. In fact, the position and definition of most symbols in the document can be well provided, which largely improves the coding experience of R. I'm excited to see how far we can go with this.

As for the information we could pre-determine in `parse_document`, it is still not clear since it depends on what functionality we wish to implement. As we consider more while moving toward xml approach I believe it would be more obvious.

renkun-ken:
@randy3k, I separate resolving package dependencies (which has significant overhead) from parse (which has little overhead). Then it seems to make sense to parse document but not resolving dependencies on document change. I test it with my large project with very long scripts, and the performance still looks good.

renkun-ken:
If performance is a concern, we may make it an opt-in option to parse on document change.

randy3k:
Thank you for working on it. The xml approach is much cleaner! I love it.

I think it’s alright to not resolve dependencies on change.

randy3k:
When writing a function body, it’s also nice to have the arguments shown in the completion list.

renkun-ken:
I'm doing my work with it for several days and it works pretty well. 

Currently, `find_enclosing_scopes` does not include the top-level scope because the current `parse_expr` already provides global variables and functions using expression-approach.

To include top-level the xpath should be

```
/exprlist | //expr[(@line1 < %d or (@line1 = %d and @col1 <= %d)) and 
        (@line2 > %d or (@line2 = %d and @col2 >= %d))]
``

but we'll get duplicate global variables in the completion. To avoid duplicate items, I don't include `/exprlist` at the moment and leave the current global variable handling as it is.

As we gradually move to the xml approach, we can completely replace the existing `parse_expr` with xml handling functions that provide at least the same information. But the framework needs more consideration and it is more complicated and needs more PRs to do it.

For this PR, I think it is good to merge now for local symbols are provided in addition to global variables.

randy3k:
Thanks again for working on it. Could you fix the CRAN check error and rebase the PR to the current master (which makes review a bit easier)?

renkun-ken:
Fixed.

renkun-ken:
Is there anything else I should add?

randy3k:
LGTM for now. Let's merge it first and then evaluate.

PR #84, opened 2019-10-07
Only provide package exported objects at ::

### LINKED ISSUE #83: Package object completion should not include non-package objects
Currently, if I type `foo::`, the completion list includes not only the exported objects of package `foo`, but also objects from `global_env`. For example,

```r
x1 <- 0
x2 <- 0
ggplot2::x|
```

At `ggplot2::`, global objects still appear in the completion list, which I think does not make sense because if I choose any of them, e.g. `ggplot2::x1` does not exist. The better behavior should be only to provide package objects under such circumstances (`::` only shows package exports and `:::` only shows package exports and non-exports).

renkun-ken:
Closes #83.

At `::`, workspace symbols are excluded and only package exported objects appear in the completion so that any choice would not refer to a non-existing object.

### DIFF
--- a/R/completion.R
+++ b/R/completion.R
@@ -59,13 +59,13 @@ workspace_completion <- function(workspace, token, package = NULL, exported_only
     completions <- list()
 
     if (is.null(package)) {
-        packages <- workspace$loaded_packages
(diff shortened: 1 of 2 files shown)


randy3k:
Thanks

PR #86, opened 2019-10-12
Customizable style

### LINKED ISSUE #85: Need customizable formatting scopes
It would be very nice to allow users to customize the range/document formatting scopes at https://github.com/REditorSupport/languageserver/blob/master/R/formatting.R#L95.

```r
    # check if the selection contains complete lines
    if (character1 != 0 || character2 < ncodeunit(lastline)) {
        scope <- "line_breaks"
    } else {
        scope <- "tokens"
    }
```

For me, I only need `indentation` level of scopes under all circumstances, or otherwise my code 

```r
data <- fst::read_fst("/path/to/my/data.fst",
  columns = c("symbol", "date", "time", "pre_close", "close",
    "volume","turnover","ret"),
  as.data.table = TRUE)
```

which I think is concise and good enough will always be formatted as

```r
data <- fst::read_fst("/path/to/my/data.fst",
    columns = c(
        "symbol", "date", "time", "pre_close", "close",
        "volume", "turnover", "ret"
    ),
    as.data.table = TRUE
)
```

which I think is a bit clumsy.

We can add some options to `languageserver` such as `languageserver.partial_formatting_scope = "line_breaks"` and `langaugeserver.complete_formatting_scope = "tokens"` to allow user to customize the behavior without adding too much complexity.

renkun-ken:
Closes #85.

This PR allows user to customize formatting style by modifying `languageserver.formatting_style` option. For example, to only apply indention level of formatting scope:

```r
options(languageserver.formatting_style = function(...) {
  styler::tidyverse_style(scope = "indention", ...)
}
```

To disable assignment operator fix by default:

```r
options(languageserver.formatting_style = function(...) {
  style <- styler::tidyverse_style(...)
  style$token$force_assignment_op <- NULL
  style
}
```

### DIFF
--- a/R/formatting.R
+++ b/R/formatting.R
@@ -2,7 +2,6 @@ get_style <- function(...) {
     style <- getOption("languageserver.formatting_style")
     if (is.null(style)) {
         style <- styler::tidyverse_style(...)
-        style$token$force_assignment_op <- NULL
     } else {
         style <- style(...)
     }
@@ -15,7 +14,7 @@ get_style <- function(...) {
 #' [styler::tidyverse_style()] style.
 #'
 #' @keywords internal
-style_file <- function(path, options) {
+style_file <- function(path, style) {
     document <- readLines(path, warn = FALSE)
     if (is_rmarkdown(path)) {
         temp_file <- tempfile(fileext = ".Rmd")
@@ -24,7 +23,7 @@ style_file <- function(path, options) {
     }
     writeLines(document, temp_file)
     styler::style_file(temp_file,
-        transformers = get_style(indent_by = options$tabSize)
+        transformers = style
     )
     contents <- readLines(temp_file, warn = FALSE)
     file.remove(temp_file)
@@ -35,16 +34,14 @@ style_file <- function(path, options) {
 #' Edit code style
 #'
 #' This functions formats a list of text using [styler::style_text()] with the
-#' [styler::tidyverse_style()] style.
+#' specified style.
 #'
 #' @keywords internal
-style_text <- function(text, options, indentation = "") {
+style_text <- function(text, style, indentation = "") {
     new_text <- tryCatch(
         styler::style_text(
             text,
-            transformers = get_style(
-                indent_by = options$tabSize
-            )
+            transformers = style
         ),
         error = function(e) e
     )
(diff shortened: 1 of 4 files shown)


randy3k:
Could you check if all styles support the argument `indent_by`? Also, do all of them has "token" as in `style$token$force_assignment_op`?

renkun-ken:
If the style creating function is based on `styler::tidyverse_style(...)`, then it must support `indent_by` which is an argument, and `style$token$force_assignment_op` is there when `scope >= tokens`, and won't cause error if not.

If user choose to create own style from scratch using `styler::create_style_guide()`, then there's no  `indent_by` anymore, and there is `style$token` (created by `styler::create_style_guide()`) but probably no `style$token$force_assignment_op` anymore if user choose to name it in other way.

This works best with all styles based on `tidyverse_style(...)`. For styles created from `styler::create_style_guide()` we may need to require that `indent_by` argument be supported, or we need to work out another way to insert the indention at editing position.

randy3k:
Is there a way to check if the style is derived from `tidyverse_style` or support the `indent_by` argument?

renkun-ken:
It seems that there's no obvious way to know if a style is derived from `tidyverse_style` since it only returns a list of lists of functions and have no class attribute. And I don't there's an obvious way to know if it supports `indent_by` either. 

I think a better way to handle this is to pass `options` to `get_style` and require that user defined style function should accept `options` as information of the document, and user is free to choose to ignore it.

In this way, user defined style function looks like 

```r
options(languageserver.formatting_style = function(options) {
  styler::tidyverse_style(scope = "indention", indent_by = options$tabSize)
}
```

To disable assignment operator fix by default:

```r
options(languageserver.formatting_style = function(options) {
  style <- styler::tidyverse_style(indent_by = options$tabSize)
  style$token$force_assignment_op <- NULL
  style
}
```

randy3k:
It looks like a viable solution. Thank you for working on it.

randy3k:
Could you also add a note in README?

renkun-ken:
Sure, I've just written some note in README.

renkun-ken:
Is there anything else I should add?

randy3k:
merged c0396f2a5cc12bad39df3f74e041e6eb57c8e5d6

PR #88, opened 2019-10-15
Provide completion for language constants

### LINKED ISSUE #87: Provide completion of language constants
I notice that there is no completion for the following language constants:

```
TRUE|FALSE|NULL|NA|NA_integer_|NA_real_|NA_complex_|NA_character_|Inf|NaN
```

Other constants are in `base` namespace so they appear in the current completion.

renkun-ken:
Closes #87.


randy3k:
Perfect. Thanks.

PR #90, opened 2019-10-15
Ignore missing packages in completion and definition

### LINKED ISSUE #81: definition does not exclude `::` in strings
In the following code

```r
xml_find_all(x, "descendant::expr")
```

when I hold Command button in VSCode, it sends a definition request. When I move my cursor to `"descendant::expr"`, the request results in an error:

```
[Error - 2:31:59 PM] Request textDocument/definition failed.
  Message: {"message":"there is no package called ‘descendant’","call":{},"package":"descendant","lib.loc":{}}

  Code: -32603 
```

It seems that hover definition does not exclude `::` in string literals.

renkun-ken:
Closes #81 

This PR uses `find.package(pkg, quiet = TRUE)` to detect if a package is installed in `workspace$get_namespace()` and `workspace$get_code()` so that they may return `NULL` if the requested package is not installed. And subsequent calls to them ignores the case where package is missing so that writing and hovering on an non-installed package (e.g. `abc::`) in the code does not trigger extra logging to notice user that the package is not installed (in vscode the bottom panel jumps to extension logging, which is annoying).

### DIFF
--- a/R/completion.R
+++ b/R/completion.R
@@ -67,6 +67,9 @@ workspace_completion <- function(workspace, token, package = NULL, exported_only
     if (is.null(package) || exported_only) {
         for (nsname in packages) {
             ns <- workspace$get_namespace(nsname)
+            if (is.null(ns)) {
+                next
+            }
             functs <- ns$functs[startsWith(ns$functs, token)]
             if (nsname == "_workspace_") {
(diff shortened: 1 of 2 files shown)


randy3k:
Seems reasonable.

PR #92, opened 2019-10-16
Add for symbol to scope completion

renkun-ken:
`for` symbol in enclosing scopes are added to the completion.

image


randy3k:
The same is true for function, e.g.,

```
foo <- function(x = 1, y = 3) {
   # cursor here
}

```
We should also extract the arguments `x` and `y` here.

renkun-ken:
@randy3k, please see https://github.com/REditorSupport/languageserver/pull/82 where function formals are already provided in completion.

randy3k:
I kept forgetting it. Anyway, good job!

renkun-ken:
Thanks! Next, we can draw a road map of the full migration from the expression approach to the XML approach, and it seems that more LSP functionalities can be implemented with the XML approach too.

PR #102, opened 2019-11-01
use task managers to handle parsing and diagnostics

randy3k:
this PR will decouple parsing and diagnostics so we don't have to wait for `lintr::lint` when parsing the document.


renkun-ken:
I tried the latest master and find the lintr option is no longer respected since I disable lintr but it still runs linting.

randy3k:
Thank you for catching it. 0bcff11 should fix it.

renkun-ken:
Thanks for the quick fix!

PR #104, opened 2019-11-03
refactor definition code

randy3k:
additionally, hovering on a user-defined function will show its signature if failed to get help.

The next step is to merge the `DefinitionCache` and `Namespace` because they share some similar structures.


randy3k:
This PR has not finished yet, but it's getting too big. So I think we could merge it first and continue later.

Issue #108, opened 2019-11-15
Provide argument definition from documentation in hover on function argument


renkun-ken:
The help documentation in hover message of function tokens seem to be very long. I'm wondering if it is possible to extract argument definition from the documentation and provide it in function argument completion and hover of e.g. `header=` in the following example:

```r
read.csv("test.csv", header = FALSE)
```

randy3k:
You could do it, but it requires some anti-engineering.

```
library(purrr)
topic <- "parse"
arg <- "encoding"
rd <- utils:::.getHelpFile(utils::help((topic)))
argtxt <- rd %>% keep(~attr(., "Rd_tag") == "\\arguments") %>% `[[`(1) %>% 
	keep(~attr(., "Rd_tag") == "\\item") %>% keep(~.[[1]][[1]][[1]] == arg)
argtxt[[1]][[2]] %>% map_if(~attr(., "Rd_tag") == "TEXT", ~.[[1]])
```

Almost there....

renkun-ken:
I tried 

```r
topic <- "read.csv"
arg <- "header"
```

and it seems that `.[[1]][[1]][[1]]` needs to be checked like

```r
keep(~attr(., "Rd_tag") == "\\item" && 
      length(.[[1]]) && length(.[[1]][[1]]) && length(.[[1]][[1]][[1]]) && .[[1]][[1]][[1]] == arg)
```

randy3k:
ah....we need more testings

randy3k:
and this kind of things cannot be done cleanly without `purrr`.

renkun-ken:
Yes, I didn't know there's structured objects like this for documentation. Such deeply recursive structures do not look nice but help a lot for the functionality.

randy3k:
I might be able to come up with something nicer later (but not before next week). Or do you want to play with it?

renkun-ken:
We may cache well-defined structures that are transformed from this so that we can easily pick up information from it.

renkun-ken:
I'm looking forward to your nicer idea for this. I'll spend some time looking at #107 and see what I can do about it.

renkun-ken:
The following code transforms the nested structure into a simple named list of character vectors, each argument corresponds to an element.

```r
topic <- "read.csv"
rd <- utils:::.getHelpFile(utils::help((topic)))
args <- NULL
items <- list()
for (item in rd) {
    if (attr(item, "Rd_tag") == "\\arguments") {
        args <- item
        break
[...]

PR #111, opened 2019-11-20
Add documentation text to signature and arg completion

renkun-ken:
This PR is an attempt to address #108.

A documentation extractor is implemented and function description is provided in signature, and argument description is provided in arg completion.

image

Originally, I wanted to provide `ParameterInformation` also in signature. But it's hard to determine which argument is the active parameter due to the complexities of R function argument matching.

### DIFF
--- a/R/hover.R
+++ b/R/hover.R
@@ -18,8 +18,8 @@ hover_reply <- function(id, uri, workspace, document, position) {
         logger$info("sig: ", sig)
         if (!is.null(sig)) {
             sig <- trimws(gsub("function\\s*", token_result$token, sig))
+            contents <- sprintf("```r\n%s\n```", sig)
         }
-        contents <- sprintf("```r\n%s\n```", sig)
     }
 
     if (is.null(contents)) {
--- a/R/hover.R
+++ b/R/hover.R
@@ -18,7 +18,7 @@ hover_reply <- function(id, uri, workspace, document, position) {
         logger$info("sig: ", sig)
         if (!is.null(sig)) {
             sig <- trimws(gsub("function\\s*", token_result$token, sig))
-            contents <- sprintf("```r\n%s\n```", sig)
+            contents <- list(kind = "markdown", value = sprintf("```r\n%s\n```", sig))
         }
     }
 
@@ -28,9 +28,7 @@ hover_reply <- function(id, uri, workspace, document, position) {
         Response$new(
             id,
             result = list(
-                contents = list(
-                    kind = "markdown",
-                    value = contents),
+                contents = contents,
                 range = token_result$range
             )
         )
--- a/R/hover.R
+++ b/R/hover.R
@@ -18,7 +18,7 @@ hover_reply <- function(id, uri, workspace, document, position) {
         logger$info("sig: ", sig)
         if (!is.null(sig)) {
             sig <- trimws(gsub("function\\s*", token_result$token, sig))
-            contents <- list(kind = "markdown", value = sprintf("```r\n%s\n```", sig))
+            contents <- sprintf("```r\n%s\n```", sig)
         }
     }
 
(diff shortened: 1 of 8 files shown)


randy3k:
@renkun-ken I have resolved a conflict here, please pull the changes.

renkun-ken:
@randy3k Thanks!

R/workspace.R
```diff
@@ -126,7 +126,7 @@ Workspace <- R6::R6Class("Workspace",
             if (is.null(pkgname)) {
                 pkgname <- self$guess_package(topic)
             }
-            if (is.null(pkgname)) {
+            if (is.null(pkgname) || pkgname == WORKSPACE) {
```
randy3k: I was about to push the same change, you are 30 sec faster than me.

renkun-ken:
This PR looks good to me and I test it editing my scripts and no problems so far. Do you have any suggestions, @randy3k?

R/hover.R
```diff
@@ -28,7 +28,9 @@ hover_reply <- function(id, uri, workspace, document, position) {
         Response$new(
             id,
             result = list(
-                contents = contents,
+                contents = list(
+                    kind = "markdown",
+                    value = contents),
```
randy3k: `workspace$get_help` returns plain text rather than markdown.

R/signature.R
```diff
@@ -17,10 +17,17 @@ signature_reply <- function(id, uri, workspace, document, position) {
...
             sig <- trimws(gsub("function\\s*", result$token, sig))
-            SignatureInformation <- list(list(label = sig))
+            SignatureInformation <- list(list(
+                label = sig,
+                documentation = list(
+                    kind = "markdown",
+                    value = paste0(doc$description, collapse = "")
+                )
+            ))
             activeSignature <- 0
```
randy3k: Do you need to take if `doc` is NULL?

R/hover.R
randy3k: Also, some hover tests failed because of this.

renkun-ken: In VSCode, the help text looks like markdown automatically, not sure why.

image

But for short text like the function signature I have to explicitly use `list(kind = "markdown", value = "string")` make it look like code.

R/signature.R
renkun-ken: `paste0(doc$description, collapse = "")` returns `""` if `doc$description` is `NULL`, but yes, explicitly checking it will make it clear.

renkun-ken:
I'll take a look at the tests.

R/hover.R
randy3k: I doublechecked again the protocol spec, if `contents` is a string, it will be treated as MarkedString automatically. 

So
```
contents <- list(kind = "markdown", value = sprintf("```r\n%s\n```", sig))
```
could be just

```
contents <- sprintf("```r\n%s\n```", sig)
```

renkun-ken: Thanks, it's true. Got confused by the different treatment of hover (which does`, completion, and signature (which don't). Just pushed a change for this.

renkun-ken:
I added some tests to check the documentation of hover, signature and arg completion.

renkun-ken:
One more issue here: When I view a package function definition, then I hover on it to see its signature, it becomes the function call and the help text no longer appears. This occurs on current `master`.

randy3k:
Thank you. Let me have a look.

randy3k:
I understand what's going on. What the definition file of `file.path` is opened, i.e.
```
file.path <- function (..., fsep = .Platform$file.sep) 
.Internal(file.path(list(...), fsep))
```
our server mistook that it is an user defined function.

randy3k:
1960923821fb831ab47aec8d0157a12b1b1c5b5a fixes it.

renkun-ken:
Thanks for the quick fix! I've rebased my commits to latest master.

randy3k:
LGTM, let's merge and evaluate.

renkun-ken:
Thanks!

PR #112, opened 2019-11-21
Merge pull request #112 from renkun-ken/completion-item-resolve

renkun-ken:
This PR is a successor of #111 which addresses #108.

In this PR, `completionItem/resolve` is implemented so that documentation can be resolved when completion item is selected, which reduces the overhead of generating completion items.

And this allows us to expand the cases where documentation is provided. Now the documentation of package, function, non-function, lazydata, and function arguments are resolved upon request.


renkun-ken:
I'll fix the test cases for this later.

randy3k:
Awesome. Thanks.

tests/testthat/test-completion.R
```diff
@@ -96,3 +93,60 @@ test_that("Completion inside a package works", {
...
+            "basename(path", # function paraemter
+            ".Mac" # non-functon: .Machine
+        ),
+        temp_file)
+    
+    client %>% did_save(temp_file)
+    
+    result <- client %>% respond_completion(temp_file, c(0, 2))
+    items <- result$items %>% keep(~.$label == "base")
+    expect_length(items, 1)
```
randy3k: I was trying to run the tests locally and I got
```
test-completion.R:116: failure: Completion item resolve works
`items` has length 2, not length 1.
```

This is `items`, not sure why there is a duplication entry in my machine.
```
[[1]]
[[1]]$label
[1] "base"

[[1]]$kind
[1] 9

[[1]]$data
[[1]]$data$type
[1] "package"



[[2]]
[[2]]$label
[1] "base"

[[2]]$kind
[1] 9

[[2]]$data
[[2]]$data$type
[1] "package"
```

Strangely it only happens on my computer, not on the CI's...I will dig deeper to understand the cause.

randy3k: I confirm that it is an issue of my end. To be more specific, it is due to a bad interaction between `packrat` and `callr`.

```
callr::r(function() .libPaths())
```
in a packrat environment includes path to system library.

renkun-ken: So is there anything we can do about it?

randy3k: I am filing a bug report to `callr` to see if they could fix it. Otherwise, I guess we could basically ignore it. (I will probably modify the test a little bit so that it won't compliant on my setup.)

PR #114, opened 2019-11-22
Provide signature for primitive functions

renkun-ken:
Since primitive functions do have `args()` output, we can provide their signature.


randy3k:
I forget why I put the check there, maybe because most primitive functions only have one argument.

renkun-ken:
Primitive functions like `sum`, `max` also have `na.rm=` argument. Since there are no formals for them, providing signature and documentation will help.

randy3k:
> Since there are no formals for them
it sounds familiar, it may be the reason I put this check here while I was working on `get_formals`. This PR LGTM.

renkun-ken:
Thanks!

PR #115, opened 2019-11-22
Check args in get_signature

renkun-ken:
Primitive functions like `function` itself has `NULL` args. No need to provide signature for such functions.


randy3k:
Ah, it’s the exact reason why I disabled primitive functions. Horray

renkun-ken:
It looks like `format(function or expression)` works, so we don't have to use `caputure.output(print(...))` which has more overhead.

randy3k:
great.

PR #118, opened 2019-11-23
Check missing e in parse_expr

### LINKED ISSUE #117: Missing argument in if causes parse_expr to fail
The following source code

```r
if (data[, sum(x) > 0]) {
  stop("error")
}
```

causes `parse_expr` to fail with following error:

```
Error in parse_expr(expr, env) : argument "e" is missing, with no default
```

renkun-ken:
Closes #117.


randy3k:
Thanks

Issue #58, opened 2019-09-18
Autocompletion per document or for the whole workspace


renkun-ken:
The current implementation of symbol provider is workspace-wide, i.e., if you open and edit 10 R script files, then all documents are parsed and all symbols are shared in the completion list when I edit one file.

This can sometimes be annoying if I need to edit a simple script and a very long script that defines a lot of symbols at the same time.

I'm wondering if it makes sense to provide symbols per document, or according to referenced documents (as mentioned in #20).

randy3k:
I guess we will need to check if the files are coming from the same package. We may need to open a [workspace](https://github.com/REditorSupport/languageserver/blob/master/R/workspace.R#L85) for each independent file.

renkun-ken:
Global namespace no longer infinitely grows with #123 merged. So the behavior is good enough for now.

PR #123, opened 2019-12-03
Store parse_data separately and combine them in GlobalNameSpace

renkun-ken:
This PR addresses #58 by modifying the mechanism of populating the global namespace. One extreme is to infinitely expand global namespace (current approach), the other extreme is to provide completion per document. This PR is an attempt to strike a balance so that global namespace is not infinitely expanding but still shares symbols between multiple documents.

The current mechanism merges objects from parse data on each workspace update. Even if a symbol is removed from all documents, the symbol still lives in the global namespace.

Storing `parse_data` of each `uri` separately and combine them on each update will remove symbols that are removed in documents, making global namespace consistent with documents.

### DIFF
--- a/R/namespace.R
+++ b/R/namespace.R
@@ -126,12 +126,16 @@ GlobalNameSpace <- R6::R6Class("GlobalNameSpace",
             NULL
         },
 
-        update = function(nonfuncts, functs, signatures, formals) {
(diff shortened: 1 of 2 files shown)


randy3k:
LGTM. Thank you for working on it.

PR #128, opened 2019-12-13
treat square and curly brackets as parens

### LINKED ISSUE #126: arg completion should not be provided in non-parentheses scope
image

randy3k:
@renkun-ken would you check if this is sufficient to fix #126?

I will write some tests later today.


renkun-ken:
It works for the particular cases in #126 but the change regard `{}`, `[]` and `{}` as the same, leading to mistakenly regarding `f{}` and `f[]` both as function calls so that arg completion are provided in these two cases.

randy3k:
Now, we check which bracket is unbalanced.

renkun-ken:
Thanks! This fixes #126 and the function call detection works well too.

Issue #129, opened 2019-12-13
incorrect highlight for variables after $


randy3k:
<img width="237" alt="Screen Shot 2019-12-12 at 8 43 40 PM" src="https://user-images.githubusercontent.com/1690993/70769889-19f70700-1d20-11ea-8f69-05180221c9d5.png">

In general, it is impossible to infer what could go after $. So I guess we should just ignore everything after $.

renkun-ken:
Do you mean document highlight? Currently, the document highlight is designed to show all tokens same with one at cursor (instead of all occurrences which might partially match), it does not try to show semantically the same token (which is not easy I believe).

randy3k:
I meant when the variable `bar` is selected, the `bar` in `foo$bar` should not be highlighted because `bar` and `foo$bar` are semantically different.

renkun-ken:
Currently, it does not do or say anything about semantics because when it comes to semantics, it can quickly get much complicated. 

The currently supported use case is to make it easier to search tokens rather than texts.

Do you prefer we go deeper in semantics? For example, when I choose `bar` I can pick out all `bar` not following `$` or `@`. If I choose `bar` in `foo$bar`, do we ignore it and stop providing document highlight? Personally, my main use case here is to search token, then I don't have a way to find all tokens named `bar`.

If it is too easy to get false positives and omit real results if we provide semantic highlight, then it looks like it's too unreliable, compared to token highlight which has simple and determined behavior which mainly serves as a token searcher.

renkun-ken:
When we consider semantics, we need to define the highlight behavior for the following cases:

On each token, what other tokens do we need to highlight?

```r
foo <- list(bar = 1)
bar <- list(foo = 1)
foo$bar + bar$foo
```

randy3k:
In general, I feel like we should avoid false positives at much as possible. I don’t mind having a few negatives due to semantics.

The current implementation is just like highlighting all the matches and we don’t even need a language server to find these matches. A regular editor plugin could do the same task more efficiently.

I think we should at least

randy3k:
Oh, the definition code have the same issue:
```
foo$bar <- 1
bar
```
`bar` will goes to `foo$bar`.

renkun-ken:
Good finding. Let me take a look at the definition code and hopefully I can make highlight more semantic.

renkun-ken:
To provide a more semantic highlight, we need to consider what we should highlight when the cursor is moved to each token in the following code:

```r
[...]

PR #130, opened 2019-12-13
Fix XPaths to detect assignments

renkun-ken:
This PR fixes the issue found at https://github.com/REditorSupport/languageserver/issues/129#issuecomment-565307510.

Now `bar` in the following code can be correctly identified in hover, definition and completion.

```r
bar <- 0
foo$bar <- 1
foo$bar = 1
bar
```


randy3k:
Could we somehow consolidate the code? It looks like they are exactly (or almost) the same.

renkun-ken:
I would love to unify them somehow but the exact XML node we need to extract in each case may be slightly different. Sometimes we need `expr`, and others we need `SYMBOL` in the `expr`. We may leave them as it is for now and come back when we have a more unified approach to work with XML parse data.

Issue #127, opened 2019-12-12
Rare error that may cause lsp to crash


renkun-ken:
When I open `R/utils.R`, the following error occurs and lsp crashed. But I cannot reproduce this error at the moment.

```
received:  Content-Length: 9352
error handling json:  {"message":"lexical error: invalid char in json text.\n             } else if (tag 464288n","call":{}}
received:  \\\\ifelse\", \"USERMACRO\")) {\n            \"\"\n        } else if (is.character(item)) {\n            trimws(item)\n        } else if (length(item)) {\n            convert_doc_to_markdown(item)\n        }\n    }))\n}\n\nconvert_doc_string <- function(doc) {\n    paste0(convert_doc_to_markdown(doc), collapse = \" \")\n}\n\nxdoc_find_enclosing_scopes <- function(x, line, col, top = FALSE) {\n    if (top) {\n        xpath <- glue(\"/exprlist | //expr[(@line1 < {line} or (@line1 = {line} and @col1 <= {col})) and\n                (@line2 > {line} or (@line2 = {line} and @col2 >= {col}))]\")\n    } else {\n        xpath <- glue(\"//expr[(@line1 < {line} or (@line1 = {line} and @col1 <= {col})) and\n                (@line2 > {line} or (@line2 = {line} and @col2 >= {col}))]\")\n    }\n    xml_find_all(x, xpath)\n}\n\nxdoc_find_token <- function(x, line, col) {\n    xpath <- glue(\"//*[not(*)]\n    [(@line1 < {line} or (@line1 = {line} and @col1 <= {col})) and\n                (@line2 > {line} or (@line2 = {line} and @col2 >= {col}))]\")\n    xml_find_first(x, xpath)\n}\n\nxml_single_quote <- function(x) {\n    x <- gsub(\"'\", \"&apos;\", x, fixed = TRUE)\n    x\n}\n"}}}Content-Length: 161
Error in self$read_header() : Unexpected non-empty line

[]
exiting
[Info  - 12:13:07 AM] Connection to server got closed. Server will restart.
[Error - 12:13:07 AM] Request textDocument/documentSymbol failed.
Error: Connection got disposed.
	at Object.dispose (/Users/ken/.vscode/extensions/reditorsupport.r-lsp-0.1.4/node_modules/vscode-jsonrpc/lib/main.js:876:25)
[...]

PR #138, opened 2019-12-18
handle unbalanced closing brackets

### LINKED ISSUE #135: Arg completion no longer works for multi-lined parentheses
Previously, signature and arg completion work for the following case:

```r
system.time({
}, gcFirst|)
```

where `|` is the cursor. It looks like it is caused by the changes made at #128.

When there are `{}`, `[]` or `()` positioned at two lines in the function call before the cursor, signature and arg completion no longer work.

randy3k:
Fix #135 

PS: note that we stop searching up when an empty line is reached for performance.

### DIFF
--- a/R/document.R
+++ b/R/document.R
@@ -123,7 +123,7 @@ Document <- R6::R6Class(
 
 #' Search backwards in a document content for a specific character
 #' @keywords internal
-find_unbalanced_bracket <- function(content, row, column, skip_empty_line = TRUE) {
+find_unbalanced_bracket <- function(content, row, column, skip_empty_line = FALSE) {
     .Call("find_unbalanced_bracket",
         PACKAGE = "languageserver",
         content, row, column, skip_empty_line
--- a/src/languageserver.c
+++ b/src/languageserver.c
@@ -2,7 +2,7 @@
 #include "reader.h"
 
 
-#if !defined(_WIN32) && !defined(_WIN64)
+#if !defined(_WIN32)
 
 #include <unistd.h> /* for getppid */
 
@@ -21,7 +21,7 @@ static const R_CallMethodDef CallEntries[] = {
     {"enclosed_by_quotes", (DL_FUNC) &enclosed_by_quotes, 2},
     {"stdin_read_char", (DL_FUNC) &stdin_read_char, 1},
     {"stdin_read_line", (DL_FUNC) &stdin_read_line},
-#if !defined(_WIN32) && !defined(_WIN64)
+#if !defined(_WIN32)
     {"process_is_detached", (DL_FUNC) &process_is_detached},
 #endif
     {NULL, NULL, 0}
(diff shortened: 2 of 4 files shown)


renkun-ken:
Thanks for your fix! It works for the

```r
system.time({
}, gcFirst|)
```

A much more common case for me is

```r
parallel::mclapply(1:100, function(x) {

}, mc.core = 20)
```

The typing of `(` and `{` auto inserts  `)` and `}` respectively, and when I hit Enter, an empty line is appended waiting for my typing. If I choose to specify `mc.cores = ` first before fill anything in the function, there will be an empty line there. Could we relax the way a bit so that such case can be handled?

randy3k:
Perhaps, an empty line is allowed if the line before the empty line ends with `{`?
What do you think?

renkun-ken:
More cases like 

```r
parallel::mclapply(1:100, function(x) {
  x1 <- x + 1
  # comment here
  x2 <- x1 + 2

  x3 <- x1 + x2

  # do something else

  x1 + x2 + x3
}, mc.core = 20)
```

renkun-ken:
I'm thinking that If the code is parse-able, then we can use the XML parse data to assist such call detection and it can be always correct if document is sync. Then `find_unbalanced_bracket` can be used to help with document without an XML parse data (when the document is never correctly parsed before).

renkun-ken:
I tested with `find_unbalanced_bracket(skip_empty_line=FALSE)` with large script (an inline function of more than 3000 lines), and the performance is still good. Maybe we don't have to make `skip_empty_line=TRUE` by default?

randy3k:
I am happy to set `skip_empty_line` to False. My main concern was the performance and I don't have a large file in hand to test.

randy3k:
Actually, we might need to set a hard limit to it, say 50 lines. I am a bit worried that we may get some false alarms at some point.

renkun-ken:
In my test with a script with 2000+ lines of inline function as in 

```r
mcparallel::mclapply(1:100, function(i) {
  # more than 2000+ lines of complicated code
}, mc.cores = 20)
```

and the arg completion at `mc.cores` still works well, which implies that there `find_unbalanced_bracket` works quite well. 

To test the robustness of `find_unbalanced_bracket(skip_empty_line=FALSE)`, I insert random symbols to make the file unparseable, the arg completion is still good. 

And I mess up the code inside the inline function with random imbalanced `(){}[]`, the arg completion stop working as supposed and it does not cause any error.

Therefore, I think it is robust enough to make `skip_empty_line=FALSE` by default. If I encounter any problems (many of my work scripts are very long, good testers of langaugeserver), I'll report timely.

renkun-ken:
When I'm testing with very long code in stdio mode, it seems that stdio mode does not work well with very long scripts, causing exactly the same error in #127 and crashes. I switch to TCP mode and everything works well.

randy3k:
Are you able to reproduce it? Let me see what happen if I open a large file.

renkun-ken:
I've updated #127 on this matter of editing long scripts and I'll try to provide a reproducible example on this when I get a chance.

renkun-ken:
This LGTM, thanks!

PR #162, opened 2020-01-19
do not highlight non-word tokens

randy3k:
This PR prevents non-word tokens from being highlighted. Currently, the highlight provider would highlight `(`, `<-` and many other operators.


renkun-ken:
Maybe we should use a set of allowed `token_name` to show highlight? My suggestion is

`c("SYMBOL", "SYMBOL_FORMALS", "SYMBOL_FUNCTION_CALL", "SYMBOL_PACKAGE", "FUNCTION", "NUM_CONST", "STR_CONST")`

which rules out all operators, `OP-*`, `SYMBOL_SUB`, `SLOT` which we already intend to exclude.

randy3k:
Alright, it is better than using a regular expression to filter.

PR #165, opened 2020-01-21
use writeLines to write raw data in Windows

### LINKED ISSUE #164: A hover returns "Loading..." when Shift-JIS characters are used in vscode
A hover returns "Loading..." when Shift-JIS characters are used on vscode.

I found **a small bug while using Japanese characters in vscode**, so I reported the details.
(I open same issue in R extension for vscode. [Here](https://github.com/Ikuyadeu/vscode-R/issues/194))

I used this package on vscode and found out that a hover (code completion) returns **"Loading..."** when Japanese characters (Shift-JIS) are entered, **although a hover works completely before putting the cursor on S-JIS character.**

Using this code, you could reproduce a bug above.

```
library() # A hover works well
set <- 1234 # A hover works well
set <- "abc" # A hover works well
set # Code completion works well

set <- "あいうえお" # Hovar retruns "Loaging..."
library() # Hovar retruns "Loaging...", too.
```

I confirmed setting.json and tried re-installing extensions and relevant package, but the problem was not improved.

My env is as below.

```
> sessionInfo()
R version 3.6.2 (2019-12-12)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 17134)

Matrix products: default

locale:
[1] LC_COLLATE=Japanese_Japan.932  LC_CTYPE=Japanese_Japan.932
[3] LC_MONETARY=Japanese_Japan.932 LC_NUMERIC=C
[5] LC_TIME=Japanese_Japan.932

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base


loaded via a namespace (and not attached):
[1] compiler_3.6.2
```

More details, I use vscode-r version 1.2., R-LSP client version 0.1.4 and latest version of languageserver (in CRAN).

How can I solve this problem?

(This is my first issue report. I'm sorry if I could not give the necessary information.)

In my other device, it is macOS, I could not reproduce this bug. From the above, it seems to be particular to Windows 10 and not UTF-8 characters.

How can I solve this problem?

randy3k:
close #164

Screen Shot 2020-01-20 at 9 24 14 PM

### DIFF
--- a/R/languageserver.R
+++ b/R/languageserver.R
@@ -108,7 +108,11 @@ LanguageServer <- R6::R6Class("LanguageServer",
         },
 
         write_text = function(text) {
-            cat(text, file = self$outputcon)
+            if (.Platform$OS.type == "windows") {
+                writeLines(text, self$outputcon, sep = "", useBytes = TRUE)
+            } else  {
+                cat(text, file = self$outputcon)
+            }
         },
 
         read_line = function() {


randy3k:
There is a related bug when the Japanese string is highlighted. Unfortunately, it is an R bug and I have reported the bug to R bugzilla: https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17696

PR #171, opened 2020-01-31
Implement DocumentLinkProvider

renkun-ken:
This PR implements `documentLinkProvider` using `STR_CONST` tokens provided by XML parsed document.

image


randy3k:
Should the link be relative to the document rather than relative to the root?

renkun-ken:
If user follows how RStudio organizes an R project, then all scripts should be run from the workspace root rather than the directory that contains the script. It's a recommended practice from RStudio and I'm not sure how widely it is adopted by R users in general.

If many users follows a different strategy of executing R code (i.e. always starting R session from the directory that contains the script), then it would not be obvious how we know how current user runs R script in the first place.

PR #179, opened 2020-02-02
Fix handling empty input of path_from_uri and path_to_uri

### LINKED ISSUE #178: documentLinkProvider error when no absolute path present in code
If there's no absolute path present in code, documentLinkProvider will produce the following error:

```r
handling request:  textDocument/documentLink

internal error:  {"message":"missing value where TRUE/FALSE needed","call":{}}

deliver:  ["Response","Response","Message","R6"]

[Error - 9:52:52 AM] Request textDocument/documentLink failed.
  Message: {"message":"missing value where TRUE/FALSE needed","call":{}}

  Code: -32603 
```

renkun-ken:
Closes #178 

The error is caused by `path_from_uri` and `path_to_uri` not working with `character(0)` input.
Both functions are fixed by checking if input `path` or `uri` has zero length. If so, an empty character vector is returned, which is consistent with the behavior of `file.path`:

```r
file.path(NULL)
file.path(character(0))
```

both return `character(0)`.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -10,8 +10,8 @@ merge_list <- function(x, y) {
 #' Paths and uris
 #' @keywords internal
 path_from_uri <- function(uri) {
-    if (is.null(uri)) {
-        return(NULL)
+    if (length(uri) == 0) {
+        return(character())
     }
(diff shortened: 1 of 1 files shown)


randy3k:
Thanks

PR #181, opened 2020-02-02
Use tryCatchStack to caputure stack on error

### LINKED ISSUE #180: Error tracking not working
f579af1575d76be7625bbaf8a86cca7fd5a46428 adds some error tracking using `traceback()`. However, it does not work with caught error such as `try({ expr}, silent = TRUE)`.

```r
foo <- function(x) {
  x + 1
}
bar <- function(x) {
  foo(x)
}
ret <- try(bar("a"), silent = TRUE)
ret
#> [1] "Error in x + 1 : non-numeric argument to binary operator\n"
#> attr(,"class")
#> [1] "try-error"
#> attr(,"condition")
#> <simpleError in x + 1: non-numeric argument to binary operator>
traceback()
#> No traceback available
```

According to https://stat.ethz.ch/R-manual/R-devel/library/base/html/traceback.html, 

>Errors which are caught via try or tryCatch do not generate a traceback, so what is printed is the call sequence for the last uncaught error, and not necessarily for the last error. 

To record a caught error, we may try the following:

```r
ret <- tryCatch(withCallingHandlers(bar("a"), error = function(e) {
  e$stack <- sys.calls()
  stop(e)
}), error = function(e) e)
ret$stack
```

```
[[1]]
tryCatch(withCallingHandlers(bar("a"), error = function(e) {
    e$stack <- sys.calls()
    stop(e)
}), error = function(e) e)

[[2]]
tryCatchList(expr, classes, parentenv, handlers)

[[3]]
tryCatchOne(expr, names, parentenv, handlers[[1L]])

[[4]]
doTryCatch(return(expr), name, parentenv, handler)

[[5]]
withCallingHandlers(bar("a"), error = function(e) {
    e$stack <- sys.calls()
    stop(e)
})

[[6]]
bar("a")

[[7]]
foo(x)

[[8]]
x + 1

[[9]]
h(simpleError(msg, call))
```

renkun-ken:
Closes #180 

This PR:

* Implements `tryCatchStack(expr, ...)` which is an enhanced version of `tryCatch(expr, ...)` to also capture stack trace on error.
* Replace all `traceback()` calls (which do not work with caught error) with `tryCatchStack` so that the logging could properly show the calling stack on error.

Now the error logging includes a stack trace like the following:

image

It makes it much easier to track down the error.

### DIFF
--- a/R/languagebase.R
+++ b/R/languagebase.R
@@ -114,7 +114,8 @@ LanguageBase <- R6::R6Class("LanguageBase",
 
         handle_raw = function(data) {
             payload <- tryCatchStack(
-                jsonlite::fromJSON(data, simplifyVector = FALSE)
+                jsonlite::fromJSON(data, simplifyVector = FALSE),
+                error = function(e) e
             )
             if (inherits(payload, "error")) {
                 logger$error("error handling json: ", payload)
--- a/R/languagebase.R
+++ b/R/languagebase.R
@@ -113,13 +113,11 @@ LanguageBase <- R6::R6Class("LanguageBase",
         },
 
         handle_raw = function(data) {
-            payload <- tryCatch(
-                jsonlite::fromJSON(data, simplifyVector = FALSE),
-                error = function(e) e
+            payload <- tryCatchStack(
+                jsonlite::fromJSON(data, simplifyVector = FALSE)
             )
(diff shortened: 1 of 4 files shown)


randy3k:
👍

PR #182, opened 2020-02-04
Reply documentSymbol and documentLink when document parsed

### LINKED ISSUE #173: Some providers not working on document initial open
Currently, we use task manager to queue tasks such as parsing, linting, and resolving package dependencies. This makes languageserver responsive most of the time.

However, as a document is open, it triggers `textDocument/didOpen`, and some subsequent  requests such as `textDocument/documentLink` and `textDocument/documentSymbol` are sent immediately before the parsing completes. The result is, there's always no document symbol and document link after a document is open. I have to modify the document which triggers `textDocument/didChange` to also trigger `textDocument/documentSymbol` and `textDocument/documentLink` but they use an outdated XML parsed document.

I'm wondering if we could force the task manager to run in sync mode on `textDocument/didOpen` and in async mode on `textDocument/didChange`? Or do we have a better way to solve this on document initial open?

renkun-ken:
Closes #173 

This PR implements a simple mechanism that delays `textDocument/documentSymbol` and `textDocument/documentLink` to the time when document `parse_data` is updated in `parse_callback`.

`document_symbol_reply` and `document_link_reply` are refined so that it returns `NULL` if `parse_data$version` does not match `document$version` in the first place. Then the request handler will push the request id, document version, and params to the corresponding reply queue, waiting for `parse_callback` for that `document$version` is done. When `parse_callback` is called, it will handle all queued replies by replying empty response to outdated requests and recall the request handler with stored `id` and `params` for the queued reply that matches the document version.

Under such mechanism, document symbols and links are computed when the same document is parsed, and both requests will be handled timely with updated `parse_data`.

### DIFF
--- a/R/document.R
+++ b/R/document.R
@@ -283,6 +283,24 @@ parse_callback <- function(self, uri, version, parse_data) {
     logger$info("parse_callback called:", list(uri = uri, version = version))
     parse_data$version <- version
     self$workspace$update_parse_data(uri, parse_data)
+
+    pending_replies <- self$pending_replies[[uri]]
+    for (name in names(pending_replies)) {
+        queue <- pending_replies[[name]]
+        handler <- self$request_handlers[[name]]
+        while (queue$size()) {
+            item <- queue$peek()
+            if (item$version < version) {
+                self$deliver(Response$new(item$id))
+                queue$pop()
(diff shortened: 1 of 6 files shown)


randy3k:
Thanks.

PR #187, opened 2020-02-05
some lsp clients may not send textDocument/didOpen

randy3k:
for example jupyterlab

### DIFF
--- a/R/languageserver.R
+++ b/R/languageserver.R
@@ -52,7 +52,7 @@ LanguageServer <- R6::R6Class("LanguageServer",
             self$inputcon <- inputcon
             self$outputcon <- outputcon
 
-            self$document <- new.env(parent = .GlobalEnv)
+            self$documents <- new.env(parent = .GlobalEnv)
             self$workspace <- Workspace$new()
 
             self$diagnostics_task_manager <- TaskManager$new()
--- a/R/languageserver.R
+++ b/R/languageserver.R
@@ -17,7 +17,7 @@ LanguageServer <- R6::R6Class("LanguageServer",
         outputcon = NULL,
         exit_flag = NULL,
 
-        documents = new.env(parent = .GlobalEnv),
+        documents = NULL,
         workspace = NULL,
 
         run_lintr = TRUE,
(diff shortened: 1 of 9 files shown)


R/link.R
```diff
@@ -5,7 +5,8 @@ document_link_reply <- function(id, uri, workspace, document, rootUri) {
     result <- NULL
 
     parse_data <- workspace$get_parse_data(uri)
-    if (is.null(parse_data) || parse_data$version != document$version) {
+    if (is.null(parse_data) ||
```
renkun-ken: Shouldn't it be

```r
is.null(parse_data) || is.null(parse_data$version) || parse_data$version != document$version
```

randy3k: I think it should be `!is.null(parse_data$version) && parse_data$version != document$version)`?

There still some edge cases that I am working on.

randy3k: I was trying to fix the error in travis and discovered that the server is not responding when version is null (which is the case for some editors, and it is also the case for our test client.).

I guess we cannot do much if we do not know the document version. Directly reply may be the best resolution!?

renkun-ken: LGTM

R/languageserver.R
```diff
@@ -52,12 +52,15 @@ LanguageServer <- R6::R6Class("LanguageServer",
             self$inputcon <- inputcon
             self$outputcon <- outputcon
 
+            self$documents <- new.env(parent = .GlobalEnv)
```
renkun-ken: I'm wondering if there is a reason that we use `environment` to store documents while we use `collections::Dict()` to store pending replies for each document uri?

randy3k: I actually developed collections after languageserver. I just didn't have the chance to move documents to Dict.

randy3k: By the way, the reason to avoid using `environment` is because `environment`'s leak memory https://github.com/r-lib/fastmap#memory-leak-examples

renkun-ken: Just recall that I read that some time ago. Thanks for your reminder.

I'm interested to move documents to Dict.

PR #189, opened 2020-02-06
Fix section symbols

### LINKED ISSUE #188: Dirty code section symbols
The current regex to extract code section symbols from comments do not use lazy matching so that the following occurs:

* In R document, `# code ######` becomes `# code ##`
* In R Markdown document, `# code ##` will be rendered as `code` but the symbol extracted is `code ##`

renkun-ken:
Closes #188 

This PR uses lazy matching (which requires `perl = TRUE`) in the regex that extract code section symbols from both R document and Rmd document.


randy3k:
Usually, I would use `stringr` functions because I don't have to care about the `perl` thing. Though, I believe that `grep` is marginally faster, so it LGTM.

PR #191, opened 2020-02-07
Support textDocument/documentColor and textDocument/colorPresentation

renkun-ken:
This PR adds support for `textDocument/documentColor` and `textDocument/colorPresentation`.

image

### DIFF
--- a/R/handlers-langfeatures.R
+++ b/R/handlers-langfeatures.R
@@ -196,7 +196,6 @@ text_document_document_color  <- function(self, id, params) {
 #' Handler to the `textDocument/colorPresentation` [Request].
 #' @keywords internal
 text_document_color_presentation  <- function(self, id, params) {
-    logger$info("text_document_color_presentation:", id, params)
     textDocument <- params$textDocument
     uri <- textDocument$uri
     document <- self$documents$get(uri)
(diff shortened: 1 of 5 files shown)


randy3k:
Wow, I never expect that `colorProvider` will get developed.

As we are enabling more providers, do we want to allow users to disable some of the providers? I don't know how other servers handle it, but I would imagine that we could read options from .Rprofile to disable specific providers.

renkun-ken:
Disabling some providers via `options(...)` in `.Rprofile` sounds perfectly okay.

R/handlers-langfeatures.R
```diff
@@ -175,15 +175,32 @@ document_link_resolve  <- function(self, id, params) {
 #' Handler to the `textDocument/documentColor` [Request].
 #' @keywords internal
 text_document_document_color  <- function(self, id, params) {
-
+    textDocument <- params$textDocument
+    uri <- textDocument$uri
+    document <- self$documents$get(uri)
+    reply <- document_color_reply(id, uri, self$workspace, document)
+    if (is.null(reply)) {
+        queue <- self$pending_replies$get(uri)[["textDocument/documentColor"]]
```
randy3k: I guess I would work on the pending mechanism at some point. The current implementation is a bit hacky.

renkun-ken: Sure, glad to see a nicer implementation :)

PR #193, opened 2020-02-07
make sure NULL converts to json's null

### LINKED ISSUE #192: lsp-mode hangs with ess-r-mode: potential issues with languageserver
This issues was originally posted at https://github.com/emacs-lsp/lsp-mode/issues/1383 and has the error log generated by `lsp-mode` associated with the issue.

I was advised to post the issue here as it seems to be a problem with `languageserver` and not the Emacs `lsp-mode` package. 

Please let me know if you need any further information.

randy3k:
close #192

### DIFF
--- a/R/protocol.R
+++ b/R/protocol.R
@@ -37,7 +37,7 @@ Request <- R6::R6Class("Request",
             if (!is.null(self$params)) {
                 payload$params <- self$params
             }
-            jsonlite::toJSON(payload, auto_unbox = TRUE, force = TRUE)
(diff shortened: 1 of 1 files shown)


randy3k:
The failures of macOS builds are irrelevant.

renkun-ken:
May I ask if returning `NULL` is sufficient for both cases (dict object expecting `{...}` and array object expecting `[...]`)?

randy3k:
I checked the lsp specification, almost all replies' results also accept `null`. So it should not be a problem.

renkun-ken:
LGTM

Issue #147, opened 2019-12-23
Remove document from workspace on didClose for non-package


renkun-ken:
Previously, #123 enables removing objects in `GlobalNameSpace` when they are removed from document. However, when more and more documents are open in editor, the workspaces contains more and more documents, so that the completion provides more and more items. Even when most documents are closed, although they are removed from `documents`, they are still left in `Workspace`, which seems a bit inconsistent.

I'm not sure if it makes sense that if the workspace is not a package, then closing a document should remove the document from the workspace so that all relevant objects of the documents are removed.

randy3k:
How about creating a `Workspace` for each individual file if the rootUri is not a package? I haven't given it a deep thought, but it seems reasonable.

Additionally, we could search for `source(...)` command to check if we need to import a workspace into another workspace.

randy3k:
With #198, all relevant objects of a file will be removed if the file is closed except if it is from an R package.

Issue #195, opened 2020-02-07
improve test coverage


randy3k:
These functions have relatively low test coverage. 

- [`formatting_reply` and `on_type_formatting_reply`](https://codecov.io/gh/REditorSupport/languageserver/src/master/R/formatting.R)
- [`document_highlight_reply`](https://codecov.io/gh/REditorSupport/languageserver/src/master/R/highlight.R)
- [`hover_reply`](https://codecov.io/gh/REditorSupport/languageserver/src/master/R/hover.R)
- [`document_link_reply `](https://codecov.io/gh/REditorSupport/languageserver/src/master/R/link.R)
- [`document_symbol_reply`](https://codecov.io/gh/REditorSupport/languageserver/src/master/R/symbol.R)

PR #198, opened 2020-02-08
Refactor Workspace

randy3k:
`documents` are now moved inside `workspace` so that they can be easily removed from the workspace.

close https://github.com/REditorSupport/languageserver/issues/147


R/completion.R
```diff
@@ -163,7 +167,8 @@ workspace_completion <- function(workspace, token, package = NULL, exported_only
                          package = package
                      ))
             })
-            nonfuncts <- ns$nonfuncts[startsWith(ns$nonfuncts, token)]
+            nonfuncts <- ns$get_completions(funct_only = FALSE, exported_only = FALSE)
```
renkun-ken: `funct_only = FALSE` sounds like not only functs but also nonfuncts. Does `funct = FALSE` have clearer meaning?

randy3k: You are right

R/namespace.R
```diff
@@ -112,94 +173,109 @@ Namespace <- R6::R6Class("Namespace",
...
+            for (doc in self$documents$values()) {
+                if (!is.null(doc$parse_data)) {
+                    if (want_functs) {
+                        completions <- c(completions, doc$parse_data$functs)
+                    } else {
+                        completions <- c(completions, doc$parse_data$nonfuncts)
+                    }
+                }
+            }
+            completions
```
randy3k: Another change is that we now generate completions on completion request rather than in the parse callback.

randy3k: Because of this, we need to decide how we should delete the document in `textDocument/didClose` for package projects and non-package project.

R/namespace.R
```diff
@@ -70,6 +90,47 @@ Namespace <- R6::R6Class("Namespace",
             formals(fn)
         },
 
+        get_documentation = function(topic) {
+            pkgname <- self$package_name
+            item <- paste0(c(pkgname, topic), collapse = "::")
```
renkun-ken: Here documentation is stored per `PackageNamespace` rather than previously global documentation, it seems that we no longer have to use `item` as key, `topic` itself is good.

R/namespace.R
```diff
@@ -1,48 +1,68 @@
...
             if (exported_only) {
-                funct %in% self$exported_functs
+                funct %in% private$exported_functs
             } else {
-                funct %in% self$functs
+                funct %in% private$functs
+            }
+        },
+
+        get_completions = function(want_functs = TRUE, exported_only = TRUE) {
```
renkun-ken: Just feel redundant about `want_functs` since any parameter in any function should mean `want` something? If you feel okay, never mind :)

randy3k: It is because in the code base `functs` is a vector of the function names and I don’t want to mix them up.

renkun-ken: Thanks for your reminder!

renkun-ken: Another question: I'm a bit confused by the name `get_completions` because `completions` means a list of `CompletionItem`s in other contexts. It only returns a character of symbols or names. Does `get_symbols` have a clearer meaning?

randy3k: True

randy3k:
@renkun-ken Any other comments before we merge it?

R/handlers-textsync.R
```diff
@@ -63,9 +63,8 @@ text_document_will_save <- function(self, params) {
 text_document_did_save <- function(self, params) {
     textDocument <- params$textDocument
     uri <- textDocument$uri
-    version <- textDocument$version
```
renkun-ken: Is there a reason we drop the document version on didSave?

randy3k: [`DidSaveTextDocumentParams`](https://microsoft.github.io/language-server-protocol/specification#textDocument_didSave) does not contain `version`.

renkun-ken: Thanks! Just notice that `DidOpenTextDocumentParams` uses `TextDocumentItem` where `version` is defined and  `DidChangeTextDocumentParams` uses `VersionedTextDocumentIdentifier` where `version` is defined.

renkun-ken:
LGTM

randy3k:
Let's merge it first. My next task is to improve test coverage #195.

renkun-ken:
Sorry, it seems we forget to refactor color provider.

PR #199, opened 2020-02-08
set O_BINARY for stdin/stdout

mattn:
On Windows, R does not read/write bytes content correctly for stdin/stdout. So content-length header is not handled for many Language Server Clients. This change set O_BINARY to stdin/stdout.

screenshot2

### DIFF
--- a/src/languageserver.c
+++ b/src/languageserver.c
@@ -1,6 +1,11 @@
 #include "search.h"
 #include "reader.h"
 
+#ifdef _WIN32
+# include <fcntl.h>
+# include <io.h>
+# include <stdio.h>
+#endif
 
 #if !defined(_WIN32)
 
@@ -29,6 +34,10 @@ static const R_CallMethodDef CallEntries[] = {
 };
 
 void R_init_languageserver(DllInfo *dll) {
+#ifdef _WIN32
+    _setmode(_fileno(stdin), _O_BINARY);
+    _setmode(_fileno(stdout), _O_BINARY);
+#endif
     R_registerRoutines(dll, NULL, CallEntries, NULL, NULL);
     R_useDynamicSymbols(dll, FALSE);
 }


randy3k:
Thank you. This is the combination of thing I did not test, namely Windows + stdio + UTF-8.

randy3k:
By the way, did you test the master branch? It contains already a fix for the UTF-8 bug.

mattn:
> By the way, did you test the master branch? It contains already a fix for the UTF-8 bug.

Yes, it seems working with utf-8 encoded files. This languageserver treat .Rprofile? Below is my .Rprofile.

```
options(encoding="UTF-8")
```

randy3k:
Yes, it reads `.Rprofile`.

I am going to close this if the master branch already fixes it. A new version of language server will be released soon. :)

mattn:
issue of utf-8 was fixed in master branch. But issue of CRLF is not fixed without this pull-request.

mattn:
Currently, this language server return responses like:

```
Content-Length: 123\r\r\n
\r\r\n
...
```

randy3k:
It seems that `\n` is translated to `\r\n` automatically.

```
r$> writeLines('\n', con = file('a.txt') , sep = "", useBytes = TRUE)

r$> read_file('a.txt')
[1] "\r\n"
```

mattn:
Yes, it is default behavior on Windows. If you want to write "\n" as is, setmode should be called or all of `file("a.txt")` should be replaced to `file("a.txt", open="rb")`.

mattn:
```
writeLines('\n', con = file('a.txt', open="wb") , sep = "", useBytes = TRUE)
```
This works fine. So stdin/stdout should be set O_BINARY.

mattn:
```
> stdout()
A connection with
description "stdout"
class       "terminal"
mode        "w"
text        "text"
opened      "opened"
can read    "no"
can write   "yes"
```
stdout is opened as text write mode

randy3k:
Thank you for looking into it. See if I would find a less intrusive way to enable binary mode of stdout. I will merge it if I wouldn't find anything better.

mattn:
If you prefer to check `isatty(stdout)` before `setmode`, let me know. I'll add code like below.

```
if (isatty(stdout)) _setmode(...)
```

Thanks.

PR #205, opened 2020-02-12
Separate resolve task from parse task

renkun-ken:
PR #204 tries to improve the performance of `parse_task` by avoiding unnecessary loading packages.

However, parsing is always fast while resolving packages with child R session is expensive. We should separate package resolving from parsing so that `parse_task` could callback quickly to provide necessary parse data for all other providers that need it and `resolve_task` should be created only when `parse_data$packages` is changed.

So the task flow looks like:

```
didOpen, didChange, didSave
* diagnostics_task -> diagnostics_callback
* parse_task -> parse_callback
  ? resolve_task -> resolve_callback
```

Another problem #204 has is that incremental resolving package could not handle reorder of package attaching, so the masking could be incorrect. The safe way to do is that whenever the library calls in the document are changed (i.e. original `parse_data$packages` changed), the resolve_task should be run. Since `resolve_task` is separated from `parse_task`, then the performance of `resolve_task` would not affect providers that use `parse_data`.

### DIFF
--- a/R/workspace.R
+++ b/R/workspace.R
@@ -173,7 +173,7 @@ Workspace <- R6::R6Class("Workspace",
         },
 
         get_parse_data = function(uri) {
-            self$documents$get(uri)$parse_data
+            self$documents$get(uri, NULL)$parse_data
         },
 
         update_loaded_packages = function() {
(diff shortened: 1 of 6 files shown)


R/workspace.R
```diff
@@ -173,7 +173,7 @@ Workspace <- R6::R6Class("Workspace",
         },
 
         get_parse_data = function(uri) {
-            self$documents$get(uri)$parse_data
+            self$documents$get(uri, NULL)$parse_data
         },
```
randy3k: If you want to avoid runtime error, you might want, `self$documents$get(uri,list())$parse_data` instead?

randy3k:
I didn't check it carefully, but it makes sense to me.

R/workspace.R
renkun-ken: In fact, `NULL$anything` is `NULL`. Do you still prefer `get(uri,list())`?

randy3k: You are right.

PR #209, opened 2020-02-13
Expression-level on-type-formatting

### LINKED ISSUE #208: Support expression-level on-type-formatting
The current on-type-formatting is an extremely simple one that only supports one-line formatting, i.e. it only tries to format the line under editing.

However, in many cases there's a need to format the current expression on `\n` or closing the brackets.

The on-type-formatting includes closing a bracket, and inserting a line break (`\n`).

The case of inserting a line break is somewhat special because the expression under editing is incomplete:

```r
mtcars %>% # hit enter
```

then the formatting should add an indention to the new line and it becomes

```r
mtcars %>%
    |
```

The case of closing brackets is typically handling complete expressions.

```r
mtcars %>%
  select(a,b, c) %>%
  mutate(x=a + b, y = b - c) %>%
  select(a, b,c)

ggplot(diamonds, aes(x = carat, y = price, color = cut)) +
  geom_point() +
  labs(title = "Scatterplot", x = "Carat", y = "Price")

if (x+1) {
  y+1
} else {
  z+1
}
```

Closing each bracket should result in formatting the current expression.

Here's my initial implementation: https://github.com/REditorSupport/languageserver/compare/master...renkun-ken:on-type-formatting. The basic approach is backward expand the range line and line from the requested point and parse the content in range, and after an expression is found, and then the next parsing results in more expressions or none (no longer parse-able) the expression under editing is found. I'm still doing experiments with it.

The trick I use to format incomplete expression with `\n` inserted is to replace the new blank line with a dot `.` so that the `.` could make the expression complete. The formatted text with last dot removed could be used as a text edit of the original text.

But there are still some edge cases such as breaking a line in the middle of an expression. I'll do more experiment with it and hopefully I could create a use-able on-type-formatter.

renkun-ken:
Closes #208 

This PR enhances the current implementation of on-type-formatting handler which only tries to format the single-line under editing.

Parsing a short chunk of code without generating source reference is super fast, which allows us to use `parse(text = ., keep.source = FALSE)` repeatedly to find the parse-able expression before current line to perform formatting.

Currently, only `)`, `]`, `}` and `\n` are used to trigger on type formatting. The closing brackets are usually with complete expressions before cursor, while the line break can be in the end of either a complete expression (a single line call like `foo()`) or an incomplete expression to be continued in the next line (typical use case is pipelines ending with ggplot's `+` or magrittr's `%>%`)

To handle the case of `\n` in the end of an incomplete expression, we use a zero strategy, i.e. if the end line is empty or only whitespace, then we replace that line with `0` to make the expression a complete one.

```r
mtcars %>% # hit enter
```

internally becomes

```
mtcars %>%
  0
```

Then we use a backward parsing strategy, i.e., we expand the code chunk backward line by line from the end line on request and try to parse the code. We stop backward parsing when 

* We have at least one expression parsed and 
* We are entering the previous expression.
  - If it is one-line expression, then we got 1 more expression.
  - If it is multi-line expression, then we got no expression

When we find the start line of the expression, and we format the code in the range and cut the zero if used.

Both backward parsing and formatting are called in `tryCatchTimeout()` which uses `setTimeout()` to limit the number of seconds to evaluate an expression. We set the timeout of backward parsing to 0.1s and formatting to 1s at the moment.

Now the on-type-formatting works nicely with the following use cases:

```r
1+1 # hit enter to format this line

f(a,b,c) # hit enter to format this line

mtcars %>% # hit enter and we get an indentation

mtcars %>%
  select(a,b, c) %>%
  mutate(x=a + b, y = b - c) %>%
  select(a, b,c) # close the bracket or hit enter

mtcars %>% select(a, b, c) # hit enter anywhere in the middle of the expression to get correct indention

if(x+1){
  y+1
}else{
  z+1
} # hit enter or close this bracket to format the whole if else expression

fun2 <- function(x,y, z) {
  x+y + z # hit enter to format this line
  x+ y +z
  local({
  local({
    x+y+z
    }) # hit enter or close bracket to format this inner local call
  }) # hit enter or close bracket to format both inner and outer local calls
} # hit enter or close bracket to format the entire function
```

Kapture 2020-02-13 at 21 02 46

### DIFF
--- a/R/document.R
+++ b/R/document.R
@@ -19,13 +19,7 @@ Document <- R6::R6Class(
 
         set_content = function(version, content) {
             self$version <- version
-            # remove last empty line
-            nline <- length(content)
-            if (nline > 0L && !nzchar(content[nline])) {
-                content <- content[-nline]
-                nline <- nline - 1
-            }
-            self$nline <- nline
+            self$nline <- length(content)
             self$content <- content
         },
 
--- a/R/utils.R
+++ b/R/utils.R
@@ -43,6 +43,14 @@ print.errorWithStack <- function(x, ...) {
     invisible(x)
 }
 
+tryCatchTimeout <- function(expr, timeout = Inf, ...) {
+    expr <- substitute(expr)
+    envir <- parent.frame()
+    setTimeLimit(timeout, transient = TRUE)
+    on.exit(setTimeLimit())
+    tryCatch(eval(expr, envir), ...)
+}
+
 capture_print <- function(x) {
     paste0(utils::capture.output(print(x)), collapse = "\n")
 }
--- a/R/formatting.R
+++ b/R/formatting.R
@@ -161,7 +161,7 @@ on_type_formatting_reply <- function(id, uri, document, point, ch, options) {
             start_line <- start_line - 1
         }
         TRUE
-    }, timeout = 0.1, error = function(e) logger$info("on_type_formatting_reply:", e))
+    }, timeout = 0.1, error = function(e) logger$info("on_type_formatting_reply:parser:", e))
 
     if (is.null(res)) {
         # timeout
@@ -192,7 +192,8 @@ on_type_formatting_reply <- function(id, uri, document, point, ch, options) {
         new_text <- tryCatchTimeout(
             style_text(content[start_line:end_line], style, indentation = indentation),
             timeout = 1,
-            error = function(e) logger$info("on_type_formatting_reply:", e))
+            error <- function(e) logger$info("on_type_formatting_reply:styler:", e)
+        )
         if (!is.null(new_text)) {
             if (use_zero) {
                 new_text <- substr(new_text, 1, nchar(new_text) - 1)
--- a/R/formatting.R
+++ b/R/formatting.R
@@ -181,11 +181,11 @@ on_type_formatting_reply <- function(id, uri, document, point, ch, options) {
             start_line <- start_line + 1
         }
 
-        logger$info("on_type_formatting_reply:", list(
-            start_line = start_line,
-            end_line = end_line,
-            chunk = content[start_line:end_line]
-        ))
+        # logger$info("on_type_formatting_reply:", list(
+        #     start_line = start_line,
+        #     end_line = end_line,
+        #     chunk = content[start_line:end_line]
+        # ))
 
         style <- get_style(options)
         indentation <- stringr::str_extract(content[start_line], "^\\s*")
--- a/R/formatting.R
+++ b/R/formatting.R
@@ -127,13 +127,16 @@ on_type_formatting_reply <- function(id, uri, document, point, ch, options) {
     end_line <- point$row + 1
     use_zero <- FALSE
     if (ch == "\n") {
+        start_line <- end_line - 1
+        if (grepl("^\\s*$", content[[start_line]])) {
+            return(Response$new(id))
+        }
         if (grepl("^\\s*(#.+)?$", content[[end_line]])) {
             # use "0" to complete the potentially incomplete expression
             last_line <- content[end_line]
             content[end_line] <- "0"
             use_zero <- TRUE
         }
-        start_line <- end_line - 1
     } else {
         start_line <- end_line
     }
--- a/R/formatting.R
+++ b/R/formatting.R
@@ -123,23 +123,76 @@ range_formatting_reply <- function(id, uri, document, range, options) {
 #' Format on type
 #' @keywords internal
 on_type_formatting_reply <- function(id, uri, document, point, ch, options) {
+    content <- document$content
+    end_line <- point$row + 1
     if (ch == "\n") {
-        row <- point$row - 1
+        # use "." to complete the partial expression
+        content[end_line] <- "."
+        start_line <- end_line - 1
     } else {
-        row <- point$row
+        start_line <- end_line
     }
-    style <- get_style(options)
-    selection <- document$line0(row)
-    indentation <- stringr::str_extract(selection, "^\\s*")
-    new_text <- style_text(selection, style, indentation = indentation)
-    if (is.null(new_text)) {
-        return(Response$new(id, list()))
+
+    # backward parse until more expressions or non-parseable
+    expr <- NULL
+    nexpr <- 0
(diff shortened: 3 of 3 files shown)


R/document.R
```diff
@@ -19,13 +19,7 @@ Document <- R6::R6Class(
...
         set_content = function(version, content) {
             self$version <- version
-            # remove last empty line
-            nline <- length(content)
-            if (nline > 0L && !nzchar(content[nline])) {
-                content <- content[-nline]
-                nline <- nline - 1
-            }
-            self$nline <- nline
+            self$nline <- length(content)
```
randy3k: I really forgot why I put this here.

randy3k: I finally recall it. It was a dirty (and incorrect) fix for our use of `stringr::str_split`.

```
r$> text = "a\nb\n"

r$> cat(text, file = "text")

r$> readr::read_lines(text)
[1] "a" "b"

r$> stringr::str_split(readr::read_file(text), "\n")
[[1]]
[1] "a" "b" ""
```

renkun-ken: Remove the last emtpy line makes it harder to handle the last line in document triggered by `\n`.

renkun-ken:
The dot strategy requires some rework because if under default styler settings, 

```r
mtcars %>%
.
```

becomes 

```r
mtcars %>%
  .()
```

R/document.R
randy3k: It shows that if you get the text from the client, there will be an additional empty line.
However, if the file is read directly from the disc, there isn't.

randy3k: I guess we need to change all the `readr::read_lines` to stringr::str_split(readr::read_file())`.

renkun-ken: or we could use `c(readr::read_lines(), "")` which looks much cheaper for global string cache?

randy3k: And the other problem is that `length(content)` would be different from the number of lines. Will it cause any issue down the road?

randy3k: That may also cause issue when `text = "a\nb"`

renkun-ken:
Another thing needs to be fixed is `\n` before comment, which also needs dot strategy but with an non-whitespace line.

Hit enter before `#`:

```r
mtcars %>% # a comment
```

R/document.R
renkun-ken: >And the other problem is that length(content) would be different from the number of lines.

I don't quite get it. Isn't number of lines be calculated from `length(content)`? If we replace `content = readr::read_lines()` with `content = c(readr::read_lines(), "")` then `length(content)` is added 1, too, which is also consistent with the number of lines in the file.

randy3k: Perhaps I didn't explain it clearly.

```
r$> text = "a\nb"

r$> cat(text, file = "foo")

r$> content = c(readr::read_lines("foo"), "")

r$> length(content)
[1] 3
```

renkun-ken: What if we don't do anything with `readr::read_lines()`? I searched all `readr::read_lines()` and it looks like nothing will be broken if we do nothing because `document$set_content` always replace content and nline together so they are always consistent. And `readr::read_lines()` from `path` will have no more number of lines than `stringr::str_split(text, "\r\n|\n")[[1]]` from `text`, which will not cause out-of-range problem.

randy3k: Then https://github.com/REditorSupport/languageserver/blob/master/R/handlers-textsync.R#L13 which is based on the text sent by the client will have different number of lines with https://github.com/REditorSupport/languageserver/blob/master/R/handlers-general.R#L35 which is based on `read_lines`.

randy3k: It may not be a problem now. I am just worried that it may cause a subtle bug down the road.

randy3k: Anyway, my issue may be non-exisiting. And I don't use on type formatting, so just merge it when you are ready.

R/utils.R
```diff
@@ -43,6 +43,14 @@ print.errorWithStack <- function(x, ...) {
     invisible(x)
 }
 
+tryCatchTimeout <- function(expr, timeout = Inf, ...) {
+    expr <- substitute(expr)
+    envir <- parent.frame()
+    setTimeLimit(timeout, transient = TRUE)
+    on.exit(setTimeLimit())
+    tryCatch(eval(expr, envir), ...)
+}
```
randy3k: Does it really work for `parse`?

renkun-ken: We don't want to interrupt inside `parse` but the while loop as whole. We just want the while loop to stop when the timeout is reached. Single parse won't take that much time even for long scripts, but when the code above the requested point is not syntactically correct, the backward parsing may go to the first line without timeout. Here timeout is used to break the while loop.

renkun-ken:
Since `.` in the end of a pipeline will be replaced with `.()` so I change it to `0` which won't be replaced and thus easier to work with.

And inserting `\n` before `#` in `mtcars %>% # a comment` is fixed too.

R/utils.R
renkun-ken: I tested with very large function and the timeout of backward parsing while loop and style_text timeout both work as expected and can be interrupted effectively.

paulofelipe:
It worked perfectly on ubuntu, but I got an error on windows:

```[Error - 12:40:44] Request textDocument/onTypeFormatting failed.
  Message: Error: unused argument (whitespace = "\\s")
Call: trimws(last_line, whitespace = "\\s")
Stack trace:
1: paste0(new_text, trimws(last_line, whitespace = "\\s"))
2: on_type_formatting_reply(id, uri, document, point, ch, options)
3: self$deliver(on_type_formatting_reply(id, uri, document, point, 
    ch, options))
4: dispatch(self, id, params)

  Code: -32603 ```

Thanks!

randy3k:
What version of R do you have?

renkun-ken:
@paulofelipe Didn't notice that `trimws` got `whitespace` since R 3.6.0. I'll use alternative ways that is compatible with older versions of R.

renkun-ken:
@paulofelipe c3290f0cc1b75849acf69583b89be5c35f8f4a88 should fix it.

paulofelipe:
It worked! Indeed I'm using and older R version (3.5.1). Thanks!

Issue #141, opened 2019-12-19
styler too slow for large script and blocks lsp


renkun-ken:
I'm editing a script of more than 1500 lines of code. I timed the formatting by `syste.time(styler::style_file(file))`:

```
ℹ 
────────────────────────────────────────
Status  Count   Legend 
✔       0       File unchanged.
ℹ       1       File changed.
✖       0       Styling threw an error.
────────────────────────────────────────
Please review the changes carefully!
   user  system elapsed 
 46.542   0.059  46.607 
```

If I execute command `Format document`, then languageserver totally is stuck at formatting the file and stops providing other functionalities, looks like the formatting task completely blocks the langaugeserver.

I also enable `formatOnPaste`, if I copy/paste a large chunk of code (hundreds of lines) to the current document, then range formatting is triggered, which also blocks languageserver.

renkun-ken:
Related: https://github.com/r-lib/styler/issues/558

randy3k:
I think `Format Document` should be a blocking action though to ensure user changes are not lost. There may be nothing we could do at this moment.

lorenzwalthert:
Also we hope that we can cache on an expression level in https://github.com/r-lib/styler/issues/570 which will speed up things even more after we implemented caching on the whole input (https://github.com/r-lib/styler/pull/538) in v1.2.0.

lorenzwalthert:
If this causes a real headache and many upvote https://github.com/r-lib/styler/issues/570, chances are that it is implemented more quickly.

lorenzwalthert:
We have now implemented https://github.com/r-lib/styler/issues/570 in the devel version of styler, meaning:

1) initial styling is as slow as before.
2) styling a file that has previously been styled (and typically only the minority of the tokens in that file changed) is now quite a bit quicker.
3) using https://github.com/lorenzwalthert/precommit will also be quicker because of 2).

Just wanted to let you know if you want to test and provide feedback. Also, we removed a bug in in the caching feature that could be detrimental (changing tokens, see https://github.com/r-lib/styler/issues/584), so if you use the GitHub version, make sure to get the latest version of styler with `remotes::install_github("r-lib/styler")`.

renkun-ken:
@lorenzwalthert  I'm using the latest styler master branch to test the caching. Is there anything I should do to enable it? I've tried styling the 1500+ line file which took more than 40 seconds. But styling it again without any modification still took nearly the same amount of time.

[...]

PR #210, opened 2020-02-16
Add instructions to enable formatting cache to README

renkun-ken:
Suggested by https://github.com/REditorSupport/languageserver/issues/141#issuecomment-584323875.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,3 +131,15 @@ options(languageserver.formatting_style = function(options) {
 ```
 
 To further customize the formatting style, please refer to [Customizing styler](https://styler.r-lib.org/articles/customizing_styler.html).
+
+### Enabling cache for formatting by styler
+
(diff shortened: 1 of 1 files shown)


randy3k:
We are now running styler in the same process as languageserver does. If a user didn't run `library(R.cache()` first, would they be using the same temporary cache path within the session?

randy3k:
If the answer is yes and given that fact that styler` imports `R.cache` already, would it be not strictly necessary for users to run `library(R.cache)` manually?
It will only be needed if a user wants persistent cache across sessions?

renkun-ken:
If a permanent directory is ever created, `R.cache` will use it; otherwise, it will use the temp directory of the session.

Yes, if the permanent directory is not created, formatting cache still works but only within current session. I'll make it clear about the difference here.

Issue #213, opened 2020-02-18
Looking at a function definition triggers lintr


intiben:
I noticed that when I open a function definition in VScode I get warnings from `lintr` about the code from that function e.g. here I looked at the `cowplot::plot_grid` function code and got:

![image](https://user-images.githubusercontent.com/56064362/74748610-15dbee00-5269-11ea-8553-a5a230fbab26.png)

Is there a way to disable this? I tried to exclude the `/tmp/` folder using the `.lintr` file but it broke the linting entirely.

renkun-ken:
It is because opening a function definition creates a temp R script that contains the function code extracted from how R prints the function. and unfortunately `lintr` under default settings does not seem to like it, e.g. when there are infix operators like `/`, lintr likes to put spaces around all infix operators as your attached image suggests while R printing does not do this.

Maybe we could disable diagnostics for files from the temp dir to resolve this.

intiben:
@renkun-ken thanks for your reply. I also came to the same conclusion as you this is why I tried to exclude `/tmp/` from lintr by adding the following to `.lintr` but lintr just stopped working then...
```
exclusions: list(list.files("/tmp", recursive = TRUE, full.names = TRUE, pattern = "\\.R$"))
```

renkun-ken:
@intiben, we are considering moving to only linting documents inside the workspace. Does it make sense to you?

intiben:
Yes that makes sense, thanks for taking my request into consideration.

renkun-ken:
We are moving to the solution that we don't lint temp files unless the workspace is started from a temp directory, and closing a file outside workspace will remove its diagnostics.

Does it make sense to you?

intiben:
Yes, thanks @renkun-ken

renkun-ken:
Closing as fixed.

PR #214, opened 2020-02-18
Disable lintr for temp files

renkun-ken:
This PR disables lintr for files from temp root as suggested in #213.

Temp root is used instead of session temp dir because an interactive session may produce a file in temp for the editor to view (e.g. vscode-R allows user to call `View(func)` in a user session to view the function code in the editor. The file is located in the temp directory of the user session rather than the session running languageserver so they have different `tempdir()` but the same temp root `dirname(tempdir())`).

### DIFF
--- a/R/handlers-textsync.R
+++ b/R/handlers-textsync.R
@@ -90,6 +90,10 @@ text_document_did_save <- function(self, params) {
 text_document_did_close <- function(self, params) {
     textDocument <- params$textDocument
     uri <- textDocument$uri
+    # remove diagnostics if file is not from workspace
+    if (!startsWith(uri, self$rootUri)) {
+        diagnostics_callback(self, uri, NULL, list())
+    }
     # do not remove document in package
     if (!(is_package(self$rootUri) && startsWith(uri, self$rootUri))) {
         self$workspace$documents$remove(uri)
--- a/R/languageserver.R
+++ b/R/languageserver.R
@@ -88,7 +88,8 @@ LanguageServer <- R6::R6Class("LanguageServer",
                 ))
             }
 
-            if (run_lintr && self$run_lintr) {
+            if (run_lintr && self$run_lintr && 
+                !startsWith(uri, path_to_uri(dirname(tempdir())))) {
                 self$diagnostics_task_manager$add_task(
                     uri,
                     diagnostics_task(self, uri, document)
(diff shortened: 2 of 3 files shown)


randy3k:
Do we want to also disable the parse task for temp files?

renkun-ken:
We want to disable lintr for temp files because their content is generated, and we are not trying to edit those files but lintr generates annoying diagnostic messages for them which are not useful.

As for parse task, I think it is still useful for a viewer to view temp files with since hover, definition, etc. are still useful here, and they do not generate annoying messages at all. Therefore, I suggest we disable lintr but keep parse task enabled.

randy3k:
It fails the lintr tests because temp files are used in the tests.

renkun-ken:
Is there a way we keep the lintr tests running but disable lintr for temp files for user?

randy3k:
Maybe we create a langauageserver specific temp directory?

renkun-ken:
Can we change the condition so that if rootUri is from temp root, then temp files should be linted. And we change the language client in lintr test cases so that it starts from the temp dir.

randy3k:
Does it actually make sense to only lint documents under rootUri?

renkun-ken:
Yes, only linting documents under rootUri makes sense to me.

Not sure how many users usually edit files outside rootUri and expect diagnostics.

If you feel okay with this, I'll make change for it.

randy3k:
Thank you for working on it, and you also need to change the  tests to reflect that.

renkun-ken:
One problem here. If lintr is disabled for files outside workspace, and if the file is edited to be syntactically incorrect, we don't have any diagnostics to tell us the file is not parse-able, which is a drawback for occasional use cases to edit files outside workspace.

randy3k:
You are right. Perhaps the simplest approach is toignore a specific temp directory created by languageserver.

renkun-ken:
I'm totally okay with ignoring a specific temp directory created by langaugeserver, though it breaks the use case where vscode-R may also create temp R file by `View(func)` in another temp dir.

One thing we could add is that we should unpublish the diagnostics when file outside workspace is closed.

randy3k:
I see. The real issue is how we could decide which temp files should be ignored and which one should not.

renkun-ken:
Therefore, the simplest approach that covers all these cases seems to be the one that ignores files from temp root if rootUri is not from temp root (i.e. we ignore temp files for non-temp workspace).

randy3k:
You convince me that it may be the best approach.

PR #217, opened 2020-02-21
Fix resolving paths

### LINKED ISSUE #215: normalizePath resolves symlinks and may cause inconsistencies
In #202, we use `normalizePath(path, "/", mustWork = FALSE)` in `path_to_uri` to ensure that the path are all absolute and force `/` as described in <https://microsoft.github.io/language-server-protocol/specifications/specification-current/#uri>.

However, according to <https://stat.ethz.ch/R-manual/R-devel/library/base/html/normalizePath.html>, it also resolves symlink to the real path on linux and macOS. This may cause some potential inconsistencies in languageserver since a workspace or a document may be the symlink to another file.

We need to review the usage of `normalizePath`, `path_from_uri`, and `path_to_uri` and see if there could be potential problems.

### LINKED ISSUE #216: Multiple links only point to the first one
Suppose there are multiple links pointing to different files in a R script:

```r
source("test1.R")
source("test2.R")
source("test3.R")
```

document link provider correctly finds 3 links but all point to the first one.

renkun-ken:
Closes #215 
Closes #216 

This PR imports `fs` to handle path-related issues.

* Replace `normalizePath()` with `fs::path_abs()` and `path.expand()` in `document_link_reply` so that symlinks are not resolved. If the link is a symlink, its target will be the symlink rather than the resolved real path.
* Add checking of `file:///` in `path_from_uri` to handle untitled document case where URI is `untitled:untitled-n`.
* Use `fs::path_has_parent` to correctly handle directory parent check since 
  * Document URI and R `tempdir()` have different cases of disk drive in Windows.
  * Document URI and rootPath may clash using `startsWith` (e.g. URI: `/home/user/project.R` and rootPath: `/home/user/project`)
* Fix document link since `path_to_uri` does not handle multiple paths.
* Add test cases for multiple document links.

Note that `path_to_uri` must respect the cases of letters or it would be inconsistent with editor generated URIs of documents.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,6 +24,7 @@ Imports:
     callr (>= 3.0.0),
     collections (>= 0.2.3),
     desc (>= 1.2.0),
+    fs (>= 1.3.1),
     jsonlite (>= 1.6),
     lintr (>= 2.0.0),
     R6 (>= 2.4.1),
@@ -37,7 +38,6 @@ Imports:
     xmlparsedata (>= 1.0.3)
 Suggests:
     covr (>= 3.4.0),
-    fs (>= 1.3.1),
     magrittr (>= 1.5),
     mockery (>= 0.4.2),
     processx (>= 3.4.1),
--- a/R/handlers-general.R
+++ b/R/handlers-general.R
@@ -23,9 +23,8 @@ on_initialize <- function(self, id, params) {
 #' @keywords internal
 on_initialized <- function(self, params) {
     logger$info("on_initialized")
-    if (is_package(self$rootUri)) {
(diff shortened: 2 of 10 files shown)


randy3k:
Could you also fix the `startsWith`
https://github.com/REditorSupport/languageserver/blob/afe3a3047a49087c97a202552d3a6e42396b3719/R/handlers-textsync.R#L98

randy3k:
I think we are ready to make a new release.

renkun-ken:
I agree.

randy3k:
Submitted, just waiting for the auto checking.

PR #223, opened 2020-02-27
Always use content in diagnose_file

### LINKED ISSUE #222: Diagnostics not working for one-line document
If the document is untitled, then `lintr::lint` will produce an error since it does not support one-line text without `\n` ending.

If the document is not untitled, then diagnostics is not performed with the latest content of document updated via text sync.

renkun-ken:
Closes #222 

This PR always uses `content` in `diagnose_file` and fixes one-line content so that it could correctly handle one-line content and untitled document.

We should avoid reading the file from the path of URI as long as we could get the content, or otherwise the unsaved or untitled document cannot be handled correctly.

### DIFF
--- a/R/diagnostics.R
+++ b/R/diagnostics.R
@@ -67,11 +67,8 @@ find_config <- function(filename) {
 #' Lint and diagnose problems in a file.
 #' @keywords internal
 diagnose_file <- function(uri, content) {
-    path <- path_from_uri(uri)
-    if (is.null(find_config(path))) {
-        linters <- getOption("languageserver.default_linters", NULL)
-    } else {
-        linters <- NULL
+    if (length(content) == 0) {
+        return(list())
     }
 
     if (is_rmarkdown(uri)) {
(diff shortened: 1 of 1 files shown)


R/diagnostics.R
```diff
@@ -74,24 +74,20 @@ diagnose_file <- function(uri, content) {
...
+            return(list())
         }
-        # use inline data
-        text <- paste0(content, collapse = "\n")
-        diagnostics <- lapply(
-            lintr::lint(text, linters = linters), diagnostic_from_lint, content = content)
+    }
 
+    if (length(content) <= 1) {
+        content <- c(content, "")
```
randy3k: smarter than what I did.

renkun-ken: Is it possible that we get `length(content) == 0`? In this case, `paste0("", collapose = "\n")` stills returns `""`.

randy3k: It seems that content is always a string, and cannot be NULL.

renkun-ken: Yes, just refined to return early for all cases with unqualified content.

Issue #125, opened 2019-12-11
Reduce the number of dependencies


nathaneastwood:
Might you consider reducing the number of dependencies `languageserver` has? For example, `readr` is used and I see that it is one of the more recent commits whereby it replaced a base R function, thus the end user needs to also install `readr` in order to use `languageserver` because it now uses `reader::read_lines` and `readr::write_lines` over the base equivalents. Similarly `glue` and `stringr` could possibly be replaced with base alternatives.

nathaneastwood:
I can always submit a PR to reduce the number of dependencies.

randy3k:
It will require some careful crafting and benchmark. For `readr`, we will get rid of it entirely became the protocol actually does not assume we have access of the local files. We should always get all the disk information from the protocol.

randy3k:
For stringr and glue, I am happy to replace them if it does not reduce readability and performance.

nathaneastwood:
Great stuff 👍🏼

PR #229, opened 2020-03-06
Remove dependency on readr

jozefhajnala:
This PR proposes to remove the dependency on `readr`. 

## Benefits

1. The full (standalone) default installation size of the `languageserver` library is around 217 MB (on linux), where the biggest dependencies are the following (in MB):

```
136     ./BH
14      ./readr
14      ./Rcpp
6       ./xml2
6       ./testthat
6       ./stringi
```

The `BH` package is however only brought in as a recursive dependency - from `readr`. Removing the dependency on `readr` therefore decreases the installation size by ~150MB, or 70%. This of makes the installation faster as well.

2. `readr` is used very rarely, so removing it has very little impact on the codebase (as seen from the diff on this PR)

## Benchmarking performance for base vs readr

For a 100kB text file (that can relate to very large source code files):

```
docker run --rm -it rocker/tidyverse R
```

## Reading

```r
tmpFile <- tempfile()
tstSmpl <- paste(sample(LETTERS, size = 35e3, replace = TRUE), collapse = ", ")
writeLines(tstSmpl, tmpFile)

microbenchmark::microbenchmark(
  times = 1000,
  readr = readr::read_lines(tmpFile),
  rbase = readLines(tmpFile)
)
```

readr is faster, but the mean difference is only ~640 _micro_ seconds:

```
# Unit: microseconds
# expr    min      lq     mean median      uq      max neval
# readr  575.3  671.35 1072.466  710.2  764.05 325187.0  1000
# rbase 1400.0 1598.60 1713.794 1659.8 1760.60   4906.9  1000
```

## Writing

readr is slighly slower (~160 _micro_ seconds):

```r
tmpFile <- tempfile()
microbenchmark::microbenchmark(
  times = 1000,
  readr = readr::write_lines(tstSmpl, tmpFile),
  rbase = writeLines(tstSmpl, tmpFile)
)
```

```
# Unit: microseconds
# expr   min     lq     mean median    uq    max neval
# readr 365.8 455.30 553.2298 504.75 574.4 6946.8  1000
# rbase 198.0 309.95 391.5435 360.35 420.9 2826.9  1000
```


randy3k:
The real reason of using readr is not it’s performance but the issue of base R functions for handling utf8 on Windows. See https://kevinushey.github.io/blog/2018/02/21/string-encoding-and-r/ for example.

jozefhajnala:
> The real reason of using readr is not it’s performance but issues of base R functions for handling utf8 on Windows. See https://kevinushey.github.io/blog/2018/02/21/string-encoding-and-r/ for example.

Thanks to the stringr import, we also have stringi available "for free" which has `stri_read_lines()` and `stri_write_lines()` that should be very good with encodings, in fact readr's `guess_encoding()` uses  `stringi::stri_enc_detect()` and others from the package as "backend".

Would you consider trying these out as replacements for readr?

randy3k:
Let's check `stri_read_lines` on Windows.

randy3k:
`stri_read_lines` and `stri_write_lines` seem to be a good replacement for readLines and writeLines.

renkun-ken:
It seems that we could also remove dependency on `stringr` and use `stringi` regex functions directly since `stringr` uses `stringi` as backend but also imports `glue` which we have already removed the dependency.

renkun-ken:
Related to #125

jozefhajnala:
I ammended the commit to use the mentioned stringi functionality, in case you want to review @renkun-ken, @randy3k.

Issue #225, opened 2020-02-28
Disable matching parenthesis in autocompletion


intiben:
I was wondering if there was any way to disable the parenthesis matching in the autocompletion feature?

Example: in the text editor type `prin` then hit tab you will get `print()` and I would like to have the option to get `print(` instead.

_Note: I am working in VScode and I am assuming that the autocompletion feature comes from languageserver not vscode-R_

renkun-ken:
Yes, this behavior comes from #168, and we could use an option to choose from the three possible behaviors: inserting plain text `f` or `f(`, or inserting snippet `f($0)`. What do you think, @randy3k ?

randy3k:
It is always unclear to me whether the server settings should be retrieved from `.Rprofile` options. While it is a canonical way to specify user settings in R, a more standard way in LSP to do so is to use `initializationOptions` or `workspace/didChangeConfiguration`.

Put aside this, I agree that .Rprofile options would be a bit approachable for general R users. I believe that we could add an option for inserting parentheses after function completions as what RStudio does.
<img width="367" alt="Screen Shot 2020-02-28 at 11 13 26 AM" src="https://user-images.githubusercontent.com/1690993/75579748-5bdc4300-5a1b-11ea-9ea3-88b8bcbc228a.png">

Though I am not sure about the auto-matching of the closing parenthesis, it seems that it is also related to one's editor settings.

By the way, it is better to display `print()` instead of `print` if the completion is going to be `print($0)`?

renkun-ken:
>It is always unclear to me whether the server settings should be retrieved from .Rprofile options. While it is a canonical way to specify user settings in R, a more standard way in LSP to do so is to use initializationOptions or workspace/didChangeConfiguration.

I prefer putting languageserver settings in `.Rprofile` because it makes more sense to apply to all clients if user has multiple editors, and some settings requires code to be provided (e.g. `languageserver.formatting_style`.

>  I believe that we could add an option for inserting parentheses after function completions as what RStudio does.

In RStudio, when `Insert parentheses after function completions` is enabled, it will insert `foo()` if `Insert matching parens/quotes` is also enabled or `foo(` if not, or otherwise it will insert plain `foo`.

If we are to add an option like `Insert parentheses after function completions`, then user could only choose from `foo` or `foo()`, without `foo(` as @intiben requested? I'm also not sure if it has anything to do with editor's auto matching parens.

[...]

PR #240, opened 2020-04-04
experiemental setting to disable snippets

randy3k:
cf: #225

### DIFF
--- a/R/completion.R
+++ b/R/completion.R
@@ -91,7 +91,7 @@ arg_completion <- function(workspace, token, funct, package = NULL, exported_onl
 #' Complete any object in the workspace
 #' @keywords internal
 workspace_completion <- function(workspace, token,
-    package = NULL, exported_only = TRUE, capabilities = NULL) {
+    package = NULL, exported_only = TRUE, snippet_support = NULL) {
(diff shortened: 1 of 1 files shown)


randy3k:
@renkun-ken any thoughts?

R/completion.R
```diff
@@ -286,6 +286,8 @@ completion_reply <- function(id, uri, workspace, document, point, capabilities)
                 items = list()
             )))
     }
+    snippet_support <- getOption("languageserver.snippet_support",
+                                 capabilities$completionItem$snippetSupport)
```
renkun-ken, line 290: If client does not have `snippetSupport` but user specifies `options(languageserver.snippet_support = TRUE)`, then there will be input like `foo($0)`. Does it make sense that `languageserver.snippet_support` should also respect client snippetSupport and should be used to disable it while client supports it rather than forcibly enable it while client does not support it? Looks like we may use something like

```r
snippet_support <- isTRUE(capabilities$completionItem$snippetSupport) &&
    getOption("languageserver.snippet_support", TRUE)
```

randy3k: good catch!

PR #246, opened 2020-04-27
Run tasks with delay to reduce CPU usage on input

### LINKED ISSUE #245: language server uses excessive CPU
CPU usage seems excessive when using `languageserver`. I've tested in both `emacs` and `vscode` and CPU usage creeps up to 90+%. This makes the experience slow and saving files takes some time. I'm not sure what other information to provide you apart from the below:

```bash
ProductName:	Mac OS X
ProductVersion:	10.14.6
BuildVersion:	18G3020

R version 3.6.3 (2020-02-29) -- "Holding the Windsock"

languageserver 0.3.5 (from github)

GNU Emacs 28.0.50

vscode 1.44.2
```

Please let me know if you need any more info.

renkun-ken:
Closes #245.

Originally, `throttle` is used to ensure that `run_tasks` is not called too frequently. However, if user input multiple characters in a short span of time (less than `t`) after some idle time, then only the first input triggers `run_tasks`, resulting in out-of-sync results. If user no longer changes the document, then the results will be out-of-sync forever.

This PR changes the throttle behavior of `TaskManager` so that each task gets a `time` at creation and the task manager only runs tasks older than `delay` seconds. Therefore, within `delay` seconds, the tasks are only replaced on user input. If any task is older than `delay`, then it will be executed. This ensures that the outcome is always sync when user stops changing the document.

Since diagnostics task, parse task are directly triggered by didChange, and resolve task is only created when parse task is finished, only diagnostics and parse need a delay to reduce unnecessary CPU usage while resolve does not need a delay.

With some trying, `delay = 0.5` seems a reasonable value for most cases which effectively drops unnecessary tasks while keeps me feeling that the lsp is responsive enough.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -27,6 +27,7 @@ Imports:
     fs (>= 1.3.1),
     jsonlite (>= 1.6),
     lintr (>= 2.0.0),
+    parallel,
     R6 (>= 2.4.1),
     repr (>= 1.1.0),
     stringi (>= 1.1.7),
--- a/R/diagnostics.R
+++ b/R/diagnostics.R
@@ -116,7 +116,7 @@ diagnostics_callback <- function(self, uri, version, diagnostics) {
 }
 
 
-diagnostics_task <- function(self, uri, document, ...) {
+diagnostics_task <- function(self, uri, document, delay = 0) {
     version <- document$version
     content <- document$content
     create_task(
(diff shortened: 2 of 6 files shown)


renkun-ken:
I change to per task delay to support the following:

1. didOpen and didClose should not delay
2. Only didChange should delay
3. Only diagnostics and parse should delay on didChange, resolve is triggered on parse complete and should not delay.

PR #248, opened 2020-05-08
respect NAMESPACE file

randy3k:
Respect NAMESPACE's `import` and `importFrom`

```
import(xml2)
importFrom(R6,R6Class)
```

It should be working now. But do not merge yet, I still want to write some tests.


renkun-ken:
Nice work!

I encounter the following error when I hover on token like `accessor` in `token_result$accessor` at `hover.R#38`:

```
[Error - 9:28:57 AM] Request textDocument/hover failed.
  Message: Error: invalid key
Call: self$imported_objects$has(object)
Stack trace:
1: self$imported_objects$has(object)
2: workspace$guess_namespace(token_result$token)
3: hover_reply(id, uri, self$workspace, document, point)
4: self$deliver(hover_reply(id, uri, self$workspace, document, point))
5: dispatch(self, id, params)

  Code: -32603 
```

renkun-ken:
It is caused by `document$detect_token()` on `bar` at `foo$bar` returns an empty token:

```
[2020-05-09 09:38:01.769] {
  "range": {
    "start": {
      "row": 29,
      "col": 53
    },
    "end": {
      "row": 29,
      "col": 53
    }
  },
  "full_token": "",
  "package": null,
  "accessor": "",
  "token": ""
}
```

but `collection::dict` does not work with empty string key.

randy3k:
This PR has a big assumption. Users do not edit NAMESPACE during a session. We are not handling the case that the file is modified dynamically. Do you want to poll the file regularly?

renkun-ken:
If we want to handle in-session changes of `NAMESPACE`, where could be the place to refresh it?

Another question is, it seems that `NAMESPACE` allows some programming in it at build time (see https://github.com/Rdatatable/data.table/blob/master/NAMESPACE). Are we going to handle such flexibility? I'm totaly okay if we don't handle it since it seems rare that `import` or `importFrom` are conditional.

randy3k:
> If we want to handle in-session changes of `NAMESPACE`, where could be the place to refresh it? Is it possible that we somehow watch the file?
> 

Maybe we could poll the file modification time in the eventloop?

https://github.com/REditorSupport/languageserver/blob/d4022f6f09197f6519a2974d57d926b25e4493e1/R/languageserver.R#L155-L165

> Another question is, it seems that `NAMESPACE` allows some programming in it at build time (see [Rdatatable/data.table:NAMESPACE@`master`](https://github.com/Rdatatable/data.table/blob/master/NAMESPACE) [](/Rdatatable/data.table/blob/HEAD@{2020-05-09T04:49:33Z}/NAMESPACE)). Are we going to handle such flexibility? I'm totaly okay if we don't handle it since it seems rare that `import` or `importFrom` are conditional.

I agree that we should pass it.

renkun-ken:
>Maybe we could poll the file modification time in the eventloop?

The event loop idea sound good to me. One thing to notice is that when user uses roxygen2 to generate `NAMESPACE`, it seems that there could be intermediate states of the file (empty or not fully generated). Therefore, we might need some error tolerance when parsing the file.

R/workspace.R
```diff
@@ -208,14 +213,72 @@ Workspace <- R6::R6Class("Workspace",
...
+                uri <- path_to_uri(path)
+                doc <- Document$new(uri, NULL, stringi::stri_read_lines(path))
+                self$documents$set(uri, doc)
+                # TODO: move text_sync to Workspace!?
+                langserver$text_sync(uri, document = doc, parse = TRUE)
+            }
+            self$import_from_NAMESPACE()
+        },
+
+        import_from_NAMESPACE = function() {
```
renkun-ken, line 231: I'm wondering if it makes sense to stick to snake_case naming convention even though the file is indeed named `NAMESPACE`. For example, we could replace all `NAMESPACE` to `namespace_file`.

Does the following look good to you?

* `import_from_namespace_file`
* `check_namespace_file`
* `namespace_file_mtime`

R/workspace.R
```diff
@@ -190,6 +211,75 @@ Workspace <- R6::R6Class("Workspace",
...
+            namespace_file_mt <- file.mtime(namespace_file)
+            if (is.na(namespace_file_mt)) {
+                return(NULL)
+            }
+            self$namespace_file_mt <- namespace_file_mt
+            exprs <- tryCatch(
+                parse(namespace_file),
+                error = function(e) list())
+            for (expr in exprs) {
+                if (expr[[1]] == "import") {
```
renkun-ken, line 245: Some package developers may prefer hand-writing `NAMESPACE` to `roxygen2`-generated one. Therefore, it is possible that the file is at an intermediate state or invalid form such as

```r
export(run)
import(xml2)
importFrom(R6,R6Class)
useDynLib(languageserver)
import
```

This will cause lsp to crash since `import` is a name which cannot be called `expr[[1]]`.

Issue #207, opened 2020-02-13
Use persistent child R sessions to process parse and diagnostics tasks


renkun-ken:
Currently, we are using `callr::r_bg()` to create background R sessions to process tasks such as `parse_task`, `diagnostics_task`, and `resolve_task`. However, the overhead of launching new process is much more than parsing the document.

Unlike `resolve_task` which requires a fresh R session to get clean search paths to resolve packages, `parse_task` and `diagnostics_task` do not need fresh R sessions to process each task. The overhead could be starting an R session and loading `languageserver` and its dependencies.

I'm wondering if we could use persistent child R sessions to process parse and diagnostics tasks. See https://github.com/r-lib/callr#persistent-r-sessions.

Following are some benchmarks:

`script.R` is a script of 200 lines.

```r
parse_test <- function() {
  file <- "script.R"
  lines <- readLines(file)
  languageserver:::parse_document(file, lines)
}

r_bg_parse <- function() {
  r <- callr::r_bg(parse_test)
  r$wait()
  r$get_result()
}

persistent_r_parse <- function() {
  # R$call(parse_test)
  # while (TRUE) {
  #   if (!is.null(R$read())) {
  #     break
  #   }
  # }
  R$run(parse_test)
}

R <- callr::r_session$new()

microbenchmark::microbenchmark(
  parse_test(),
  r_bg_parse(),
  persistent_r_parse(),
  times = 100
)
```

```
Unit: milliseconds
                 expr       min        lq      mean    median        uq        max neval
         parse_test()  19.17827  20.23432  21.94873  20.88153  21.88695   35.96008   100
         r_bg_parse() 496.95088 545.88191 735.77912 576.86343 887.01842 4612.39674   100
 persistent_r_parse() 105.28276 111.75743 115.59718 115.08366 117.73012  198.30838   100
```

The approach we are using is `r_bg_parse()`. Its performance is quite bad and unstable compared to direct `parse_test()` and `persistent_r_parse()`.

randy3k:
The task manager was originally written for lintr/diagnostic which could be quite slow. I actually am not sure if we need to run parse_document in `callr`. If parsing is quick (even for large files), we could put it in the main process.

For diagnostic tasks, it does make sense to have a persistent session per document/workspace.

renkun-ken:
[...]

PR #265, opened 2020-05-28
add session pool for task manager

qinwf:
reuse r session between `callr::r_bg`

1. added two R6 class `Session` and `SessionPool`

2. `parse task` and `resolve task` use a shared pool, `diagnostics task` use a standalone pool.

### master baseline

```
devtools::install_github("qinwf/languageserver@master-baseline")
```

compare with the master branch

screen recording [master.mov.zip](https://github.com/REditorSupport/languageserver/files/4693323/master.mov.zip)

outline display in 13s, total startup 20s

### pool

```
devtools::install_github("qinwf/languageserver@task-pool")
```

screen recording [session-pool.mov.zip](https://github.com/REditorSupport/languageserver/files/4693328/session-pool.mov.zip)

outline display in 9s, total startup 14s


randy3k:
I think @renkun-ken has been playing with a persistent session. He may know better about it.

renkun-ken:
In #207, I suggested using persistent sessions to reduce the overhead of frequent create or kill background processes, but I didn't come up with a good solution that could be safely interrupted.

@qinwf's implementation and benchmarking look quite promising. I'll take a closer look at this soon.

qinwf:
> Unlike resolve_task which requires a fresh R session to get clean search paths to resolve packages.

Just read this comment from #207.

I did not know that `resolve task` will require a session restart when the task finished.

So `resolve_task_manager` can skip using session pool and use `r_bg` directly.

renkun-ken:
@qinwf Yes you are right, `resolve_task` relies on clean R session to get clean `search()`. Would you like to change the code accordingly?

@randy3k It seems that we should use `resolve_task_manager` at https://github.com/REditorSupport/languageserver/blob/master/R/document.R#L288? Or `resolve_task_manager` is not used at all.

R/document.R
```diff
@@ -285,7 +285,7 @@ parse_callback <- function(self, uri, version, parse_data) {
 
     if (!identical(old_parse_data$packages, parse_data$packages)) {
         if (length(parse_data$packages)) {
-            self$parse_task_manager$add_task(
+            self$resolve_task_manager$add_task(
```
qinwf, line 288: updated with resolve_task_manager here

renkun-ken: looks good.

renkun-ken:
@qinwf, do you have a good way to make `resolve_task` use `r_bg()` instead of session pool?

R/languageserver.R
```diff
@@ -52,9 +52,19 @@ LanguageServer <- R6::R6Class("LanguageServer",
...
+            parse_pool <- SessionPool$new(pool_size, "parse")
+            # diagnostics is slower, so use a seperated pool
+            diagnostics_pool <- SessionPool$new(pool_size, "diagnostics")
+
+            self$parse_task_manager <- TaskManager$new("parse", parse_pool)
+            self$diagnostics_task_manager <- TaskManager$new("diagnostics", diagnostics_pool)
+
+            # no pool for resolve task
+            # resolve task require a new session for every task
+            self$resolve_task_manager <- TaskManager$new("resolve", NULL)
```
qinwf, line 67: > do you have a good way to make resolve_task use r_bg() instead of session pool?

it is using `r_bg()`

 see also https://github.com/REditorSupport/languageserver/pull/265/files#diff-8e1fe5fc7dc74b5bae6f19a701dc6157R25

R/task.R
```diff
@@ -17,20 +18,38 @@ Task <- R6::R6Class("Task",
...
-            private$process <- callr::r_bg(
-                private$target,
-                private$args,
-                system_profile = TRUE, user_profile = TRUE
-            )
+        start = function(session) {
+            # no session, use new session
+            if (is.null(session)) {
+                private$should_release <- FALSE
+                private$process <- callr::r_bg(
```
qinwf, line 25: use `r_bg()` here

R/task.R
```diff
@@ -53,33 +72,56 @@ TaskManager <- R6::R6Class("TaskManager",
...
             private$pending_tasks$set(id, task)
         },
         run_tasks = function(cpu_load = 0.5) {
-            n <- max(max(private$cpus * cpu_load, 1) - private$running_tasks$size(), 0)
+            n <- 0
+            if (private$use_session) {
+                n <- private$session_pool$get_idle_size()
+            } else {
+                # use r_bg
+                n <- max(max(private$cpus * cpu_load, 1) - private$running_tasks$size(), 0)
```
renkun-ken, line 98: It looks like `private$cpus` is never initialized here.

R/languageserver.R
renkun-ken: looks good

R/task.R
```diff
@@ -53,33 +72,56 @@ TaskManager <- R6::R6Class("TaskManager",
...
-        running_tasks = NULL
+        running_tasks = NULL,
+        session_pool = NULL,
+        name = NULL,
+        use_session = FALSE
     ),
     public = list(
-        initialize = function() {
+        initialize = function(name, session_pool = NULL) {
             private$cpus <- parallel::detectCores()
```
qinwf, line 82: `private$cpus` is never initialized here

renkun-ken: Sorry, just mixed `TaskManager` and `Task` initialization.

renkun-ken:
The code looks good to me.

I benchmark the startup time of loading the following packages:

* lintr

master: 10.766 secs
task-pool: 5.55 secs

* data.table

master: 7.225 secs
task-pool: 4.223 secs

Looks like the loading time could be almost reduced to half in general. Good work, @qinwf!
Go ahead if you think there's something more to be done (e.g. refine the comments and logging).

@randy3k Do you think if there's anything to be updated?

R/session.R
```diff
@@ -0,0 +1,243 @@
...
+            return(private$is_running)
+        },
+        get_result = function() {
+            private$result
+        },
+        restart = function(should_release = FALSE) {
+            logger$info(private$pool_name, "session restart", private$id)
+            private$session$close(grace = 100)
+            private$session <- callr::r_session$new(
+                callr::r_session_options(system_profile = TRUE, user_profile = TRUE),
```
renkun-ken, line 152: I'm wondering if we need `supervise=TRUE` here?

R/task.R
```diff
@@ -17,20 +18,38 @@ Task <- R6::R6Class("Task",
...
-                system_profile = TRUE, user_profile = TRUE
-            )
+        start = function(session) {
+            # no session, use new session
+            if (is.null(session)) {
+                private$should_release <- FALSE
+                private$process <- callr::r_bg(
+                    private$target,
+                    private$args,
+                    system_profile = TRUE, user_profile = TRUE
```
renkun-ken, line 28: do we need `supervise=TRUE` here?

R/session.R
renkun-ken: I kill the languageserver session with both `supervise=TRUE` and `supervise=FALSE` but I can't observe any difference regarding the task sessions. Killing languageserver sessions seems always kills task sessions immediately and looks like there's no intermediate state where too many task sessions exist.

Then I'm not sure when will `supervise=TRUE` be useful anyway.

renkun-ken: The implementation of superviser is https://github.com/r-lib/processx/blob/master/R/supervisor.R, looks like there's a standalone process to do this. It seems that it is only useful when the R session crashes. I think we could leave this with default `supervise=FALSE` for now unless some cases impose strong necessity.

randy3k:
@renkun-ken I remember you mentioned slow parsing time for long scripts. Could you try it with long scripts?

renkun-ken:
@randy, my long scripts are not contained in a package. The time of starting a workspace with a long script using master and task-pool is almost identical. Maybe we could benchmark with a package where there are many very long scripts?

randy3k:
I meant in #207, you mentioned that

> Actually, yesterday I implemented a version that put parse_document in main process (which is the original version of resolve-task branch in my fork, but I dropped it and replaced it with #205), and for larger document (500 - 1000 lines of code), the delay is quite notable such as there's a delay of completion, notable delay in completion resolve, etc.


Is it still relevant now? Or it has been resolved because the resolve task is now on demand only!?

renkun-ken:
@randy3k I think it is not relevant as the session pool removes the overhead of frequent creating and killing sessions, loading languageserver package and used dependencies each time, etc. Now diagnostics task and parse task are taking advantage of persistent sessions from the separate pools. And resolve task (with minimal dependency) is still run with `r_bg` on parse task callback.

renkun-ken:
I've been using this branch in work and everything looks good.

randy3k:
@qinwf 
Thanks for your contribution, we do not get contributions from others very often.

randy3k:
It seems that the windows CI failure was not an accident. It broke the CI test on Windows.

renkun-ken:
Starting from https://github.com/REditorSupport/languageserver/actions/runs/118752003 (0dca70c), the Windows CI failed.

randy3k:
Trying to reproduce it on my VM.

randy3k:
I got an untested theory. We are using `rcmdcheck::rcmdcheck` to run test and it uses `callr`'s session under the hood. And it seems that our new session pool is somehow conflicting with `rcmdcheck`. `devtools::test/check` is okay though.

randy3k:
Actually, `devtools::test` also fails 

https://github.com/REditorSupport/languageserver/runs/722483515

For some reason, it was stuck in testing document symbols.

qinwf:
I think it may be related to CPU cores and resources. 

On GitHub Actions, those VMs may have limited CPU cores and resources. it will at least start 5 r instance for a single-core machine. It may be an issue for a small VM with 512mb memory.

And they will not be able to startup.

Maybe we should use fewer sessions on one CPU core or two CPU cores machine with memory less than 512mb.

Just one session or two is enough.

qinwf:
https://github.com/REditorSupport/languageserver/blob/5d9f21a1f47a412a84260a960177cba1d33a31b6/R/languageserver.R#L55-L63

resources setting is available here

renkun-ken:
I alter the `pool_size` formula to

```r
pool_size <- min(max(floor(cpus / 2), 1), 3)
```

and the CI passed at https://github.com/renkun-ken/languageserver/actions/runs/119759587.

Let me send an PR from my branch.

PR #268, opened 2020-06-01
Fix range formatting handling initial indentation

### LINKED ISSUE #264: Range formatting always adds indention when the range has initial indention
Perform range formatting on the following code and it works well.

```r
data <- dbQuery(
  con,
  "select name, group, max(update_time) as max_time
     from users
     where date = 20200525
     group by group"
)
```

However, it does not work well when the selection has initial indention like the following:

```r
fun <- function(x, y) {
  data <- dbQuery(
    con,
    "select name, group, max(update_time) as max_time
     from users
     where date = 20200525
     group by group"
  )
}
```

image

Once:

```r
fun <- function(x, y) {
  data <- dbQuery(
    con,
    "select name, group, max(update_time) as max_time
       from users
       where date = 20200525
       group by group"
  )
}
```

Twice:

```r
fun <- function(x, y) {
  data <- dbQuery(
    con,
    "select name, group, max(update_time) as max_time
         from users
         where date = 20200525
         group by group"
  )
}
```

More times:

```r
fun <- function(x, y) {
  data <- dbQuery(
    con,
    "select name, group, max(update_time) as max_time
               from users
               where date = 20200525
               group by group"
  )
}
```

This is particular annoying if on-type-formatting triggers range formatting inside the function like above.

renkun-ken:
Closes #264.

If the selection range has non-zero indentation, then a `apply_initial_indention` function injected to the formatting styles so that the top level expressions get the indentation.

For example, to format the code inside the the following function

```r
my_fun <- function() {
  data <- query(
    con ,
    "select name, group, max(update_time) as max_time
from users
where date = ?date
group by group",
date = 20200601
  )
x = 2
y = 3  
}
```

the code will be enclosed inside `{ ... } ` and then `apply_initial_indention` will apply the indention to all top-level expressions of the selection, then the curly braces are droped to get the formatted text.

The result is like 

```r
my_fun <- function() {
  data <- query(
    con,
    "select name, group, max(update_time) as max_time
from users
where date = ?date
group by group",
    date = 20200601
  )
  x = 2
  y = 3
}
```

### DIFF
--- a/R/formatting.R
+++ b/R/formatting.R
@@ -14,19 +14,35 @@ get_style <- function(options) {
 #' specified style.
 #'
 #' @keywords internal
-style_text <- function(text, style, indentation = "") {
-    new_text <- tryCatch(
-        styler::style_text(
-            text,
-            transformers = style
-        ),
-        error = function(e) e
-    )
+style_text <- function(text, style, indentation = 0L) {
+    if (indentation == 0L) {
+        new_text <- tryCatch(
+            styler::style_text(
+                text,
+                transformers = style
+            ),
+            error = function(e) e
+        )
+    } else {
+        style$indention$apply_initial_indention <- function(pd_nested) {
+            if (2L %in% pd_nested$pos_id) {
+                pd_nested[pd_nested$token == "expr", "indent"] <- indentation
+            }
+            pd_nested
+        }
+        new_text <- tryCatch({
+            out <- styler::style_text(
+                c("{", text, "}"),
+                transformers = style
+            )
+            out <- out[-c(1, length(out))]
+        }, error = function(e) e)
+    }
     if (inherits(new_text, "error")) {
         logger$info("formatting error:", new_text$message)
         return(NULL)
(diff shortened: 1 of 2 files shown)


R/formatting.R
```diff
@@ -106,7 +122,7 @@ range_formatting_reply <- function(id, uri, document, range, options) {
     }
 
     selection <- document$content[(row1:row2) + 1]
-    indentation <- stringi::stri_extract_first_regex(selection[1], "^\\s*")
+    indentation <- nchar(stringi::stri_extract_first_regex(selection[1], "^\\s*"))
     new_text <- style_text(selection, style, indentation = indentation)
```
randy3k, line 126: it doesn't work with tabs unfortunately. Though it seems that styler doesn't support tabs https://github.com/r-lib/styler/issues/347

renkun-ken: Yes, it does not support tabs. So I just ignore this case unless somebody complains.

Issue #242, opened 2020-04-19
R 4.0 raw string support


renkun-ken:
We might need to make some changes to string-related providers so that raw strings introduced in R 4.0 could be handled correctly (see https://stat.ethz.ch/R-manual/R-devel/library/base/html/Quotes.html).

For example, documentLinkProvider assumes all strings are in the form of `"str"` and `'str'` provided by xmlparsedata. We'll take a look at the output of parse and parse data so that we could handle the raw strings accordingly.

randy3k:
`search.c` also needs some serious refactoring too.

randy3k:
hopefully, the new syntax won't be adopted very quickly by users. By the way, radian v0.5 is out and it supports the new raw string syntax.

renkun-ken:
We need to rely on xmlparsedata output to find string literals in the document but it seems that the package has some issue (https://github.com/r-lib/xmlparsedata/issues/10) with R 4.0 raw strings too as I played around.

renkun-ken:
As R 4.0.1 fixes

>Parse data for raw strings is now recorded correctly. Reported by Gabor Csardi. 

The latest parse data of the following expression

```r
x <- r"(hello, "world")"
```

becomes 

```
> getParseData(e)                                                              
  line1 col1 line2 col2 id parent       token terminal                text
7     1    1     1   24  7      0        expr    FALSE                    
1     1    1     1    1  1      3      SYMBOL     TRUE                   x
3     1    1     1    1  3      7        expr    FALSE                    
2     1    3     1    4  2      7 LEFT_ASSIGN     TRUE                  <-
4     1    6     1   24  4      6   STR_CONST     TRUE r"(hello, "world")"
6     1    6     1   24  6      7        expr    FALSE                    
```

The trimming one character from both beginning and end no longer works. We have to extract the string text using other methods.

renkun-ken:
In all cases, we could use `parse(text = all_strings)` the get the string texts correctly, which has exactly the same support of raw strings as the R version supports.

renkun-ken:
Closing as resolved.

PR #274, opened 2020-06-10
Fix handling raw string in link and color

renkun-ken:
Related #242 

This PR

* Fixes raw string handling in document link using `col1` and `col2` attributes of  `STR_CONST` (which are correct) and `parse()` to extract the right text.
* Ignores raw string in color provider to avoid incorrect replacement of color from color picker.
* Adds test cases of handling raw strings in document link.


randy3k:
LGTM. Thanks for working on this.

PR #276, opened 2020-06-13
support R 4.0 raw string search

randy3k:
TODO:

- [x] write tests
- [ ] ~~better multiline support~~


src/fsm.c
```diff
@@ -6,24 +6,106 @@ typedef struct {
...
+                state->double_quoted = 1;
+                state->raw_dashes_temp = 0;
+                return;
+            } else {
+                state->raw_parse_state = 0;
+                state->double_quoted = 0;
+                state->single_quoted = 0;
+            }
+        } else if (state->raw_parse_state == 2) {
+            if (c == '(' || c == '[' || c == '{') {
```
renkun-ken, line 42: Looks like `|` works too. 

See https://github.com/wch/r-source/blob/1977750ab278e6e3ea5c29b6a1916c7398e9e2fa/src/main/gram.c#L5038.

randy3k:
I didn't have time to work on this, so I guess we will skip it for v0.3.7

src/fsm.c
randy3k: Do you think we should also support `|`?

randy3k:
I have decided not to improve multiline support in this PR but rather to merge it first.

After merging this PR, I reckon we should freeze the master branch and prepare a release.

src/fsm.c
renkun-ken: I guess we should not since it is undocumented behavior.

Issue #277, opened 2020-06-16
Error: key not found


gitbobo:
System:window 10,vs code 1.46.0.R lsp client 0.16.0,R 3.5.3

[Error - 10:41:16] Request textDocument/colorPresentation failed.
  Message: Error: key not found
Call: self$workspace$documents$get(uri)
Stack trace:
1: self$workspace$documents$get(uri)
2: dispatch(self, id, params)

  Code: -32603 
[2020-06-16 10:41:21.121] received:  Content-Length: 234
[2020-06-16 10:41:21.133] received payload.
[2020-06-16 10:41:21.133] handling request:  textDocument/documentHighlight
[2020-06-16 10:41:21.133] internal error: Error: key not found
Call: self$workspace$documents$get(uri)
Stack trace:
1: self$workspace$documents$get(uri)
2: dispatch(self, id, params)
[2020-06-16 10:41:21.134] deliver:  ["Response", "Response", "Message", "R6"]

[Error - 10:41:21] Request textDocument/documentHighlight failed.
  Message: Error: key not found
Call: self$workspace$documents$get(uri)
Stack trace:
1: self$workspace$documents$get(uri)
2: dispatch(self, id, params)

  Code: -32603

renkun-ken:
Are you using the latest development version of languageserve at master branch or the CRAN version?

gitbobo:
> 
> 
> Are you using the latest development version of languageserve at master branch or the CRAN version?
> 
> Would you like to try the latest development version and see if you could reproduce this?

languageserver installed version 0.3.5,is that latest version?

renkun-ken:
Would you like to install the latest development version via

```r
remotes::install_github("REditorSupport/languageserver")
```

and see if you still have this problem?

randy3k:
that's odd. Do the file name and the file path contain any special characters?

gitbobo:
> 
> 
> that's odd. Do the file name and the file path contain any special characters?
Thank's for your remind,it's really because of one special character.

randy3k:
Mind telling us what it was?

gitbobo:
> 
> 
> Mind telling us what it was?
setwd("E:/collections/river side middle school students test record 2019-4-22次/832")
When I remove Chinese characters,everything going good.  
In Rstudio,there's no such a problem.

randy3k:
Thanks. There may be an encoding problem somewhere.

randy3k:
Actually, I see a different error
```

[...]

PR #278, opened 2020-06-16
make sure the path is UTF-8

randy3k:
it should fix some of the issues in #277.

We still need to fix the round trip issue

```
r$> a = "file:///foo/bar/中文.R"

r$> (b = languageserver:::path_from_uri(a))
[1] "/foo/bar/中文.R"

r$> (languageserver:::path_to_uri(b))
[1] "file:///foo/bar/%E4%B8%AD%E6%96%87.R"
```

### DIFF
--- a/R/handlers-general.R
+++ b/R/handlers-general.R
@@ -6,7 +6,7 @@
 on_initialize <- function(self, id, params) {
     logger$info("initialization config: ", params)
     self$processId <- params$processId
-    self$rootUri <- params$rootUri
+    self$rootUri <- utils::URLencode(params$rootUri)
     self$rootPath <- path_from_uri(self$rootUri)
     self$workspace <- Workspace$new(self$rootPath)
     self$initializationOptions <- params$initializationOptions
--- a/R/handlers-general.R
+++ b/R/handlers-general.R
@@ -6,7 +6,7 @@
 on_initialize <- function(self, id, params) {
     logger$info("initialization config: ", params)
     self$processId <- params$processId
-    self$rootUri <- utils::URLencode(params$rootUri)
+    self$rootUri <- uri_escape_unicode(params$rootUri)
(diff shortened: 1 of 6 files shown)


randy3k:
One solution is to always escape the uri from `textDocument$uri`. Say for example,

```
uri <- URLencode(textDocument$uri)
```

renkun-ken:
Do you mean always `URLencode` the incoming URI from editor and `URLdecode` the outcoming URI to editor? I think it should work with VSCode but if an editor uses encoded URI that escapes all unicode characters, then returning a decoded URI seems undesired by such editors?

renkun-ken:
I don't know what other editors do but VSCode seems to only encode a minimal set of reserved characters. For example, on macOS:

```
测试 copy !$&'()*+,;=:?@#.R
```

becomes

```
测试%20copy%20!$&%27()*+,;=:%3F@%23.R
```

randy3k:
>  `URLdecode` the outcoming URI to editor? 
No. We still need to escape special characters like spaces.

But it raises another issue, would vscode understands both `"测试%20copy%20!$&%27()*+,;=:%3F@%23.R"` and `"%E6%B5%8B%E8%AF%95%20copy%20!$&'()*+,;=:?@#.R"`?

randy3k:
Oops, 

I just notice that vscode escapes `'` but R does not so `URLencode` might not work.

renkun-ken:
I tried the latest commit in this PR and I don't observe any problem working with `测试 copy !$&'()*+,;=:?@#.R`.

renkun-ken:
Always encoding incoming URIs and not decoding outcoming URIs assumes the editor always does decoding, which I think is quite a safe assumption.

randy3k:
It also fixes #279 (but I don't understand why).

renkun-ken:
Thanks for working on this. LGTM.

PR #284, opened 2020-06-25
Use lint text if supported

### LINKED ISSUE #80: Provide diagnostics on non-installed packages
The current diagnotics use `lintr` which is mainly focused on style diagnostics. However, code correctness is largely not diagnosed. An obvious one is to check if there is any `library(pkg)` call that points to non-installed package. Similar calls include `require(pkg)`, `loadNamespace("pkg")`, `requireNamespace("pkg")`, `pkg::foo`, `pkg:::bar`.

The recently released [`lintr v2.0.0`](https://www.jimhester.com/post/lintr-2-0-0/) supports [customized lintr](https://cran.r-project.org/web/packages/lintr/vignettes/creating_linters.html). I'm not sure if it is better to use this or separate correctness checks from linting.

### LINKED ISSUE #235: .lintr defaults are not respected
I have a `.lintr` file in my repo root and running `lintr::lint_package()` works fine.

However, the disabled linters are not honored from languageserver and still appear in the problems pane.

`.lintr`

```
linters: with_defaults(assignment_linter = NULL, object_name_linter = NULL)
```

### LINKED ISSUE #238: Packages not loaded prior to linting with coc-r-lsp
Not sure if this is an issue of this package, or a problem of [coc-r-lsp](https://github.com/neoclide/coc-r-lsp), but I'm encountering some difficulties setting them up to work together.

Although linting basically works out of the box, it looks like the current buffer is getting processed by `lintr` without loading the package (or any dependencies, imports, ...) first. This results in many warnings generated by the `object_usage_linter`. If I run `languageserver::diagnose_file("<file>", readLines("<file>")` everything looks good OTOH.

### LINKED ISSUE #282: Exclusions in .lintr not working
From https://github.com/Ikuyadeu/vscode-R/issues/354.

The `exclusions` in `.lintr` is not working:

`.lintr`:

```
exclusions: list("test1.R")
```

`test1.R`:

```r
1+1
```

The file is still diagnosed, implying that the `exclusions` is not respected.

This is caused by that we lint a temp file with potential unsaved edits rather the original file in the workspace. The exclusions do not apply to the temp file path.

renkun-ken:
Close #80
Close #235 
Close #238 
Close #282 

https://github.com/jimhester/lintr/pull/503 adds a `text=` argument to `lintr::lint` to make it easier for code editors to lint a text as if it is from a filename so that lint config is kept effective.


randy3k:
LGTM except you might need to `redocument`. Actually do we really want to document all the internal functions? If they are not needed, perhaps it is easier to remove those documentations.

renkun-ken:
Since https://github.com/jimhester/lintr/pull/503 is merged, we could wait for the latest lintr (probably 3.0.0) to be released, then we could import `lintr (>= 3.0.0)` and this PR could be merged.

renkun-ken:
@randy3k  Since we have support for both old lintr (current release) and latest lintr with `lint(text=)` by checking if the `text` argument exists, I guess we could merge this without having to rely on lintr >= 3.0.0?

R/diagnostics.R
```diff
@@ -79,27 +79,35 @@ diagnose_file <- function(uri, content) {
...
 
     path <- path_from_uri(uri)
+
+    if (length(content) == 1) {
+        content <- c(content, "")
+    }
+
     linter_file <- find_config(path)
+    linters <- NULL
+    lint_text_support <- "text" %in% names(formals(lintr::lint))
```
randy3k, line 89: Perhaps we should add a note here about the future release of lintr.

randy3k (approved):
I am a bit hesitated to merge it. With that say, I guess it is an acceptable solution for now, just need to put a note here to remind ourselves in the future.

renkun-ken:
The tests only works with latest lintr rather than the current CRAN release. I guess we should postpone this PR until lintr 3.0.0 is released and then we could directly import lintr 3.0.0 and do not have to detect `text=` like this and those test cases should work too.

renkun-ken:
I guess we could merge this PR into master now. The development version of `lintr` is required until lintr 3.0.0 is released to CRAN.

R/diagnostics.R
```diff
@@ -79,27 +79,39 @@ diagnose_file <- function(uri, content) {
...
+        content <- c(content, "")
+    }
+
     linter_file <- find_config(path)
+    linters <- NULL
+
+    # In the upcoming lintr 3.0.0, lint(text=) is supported so that
+    # we could supply both filename and text and the text will be
+    # regarded as the content of the file.
+    lint_text_support <- "text" %in% names(formals(lintr::lint))
```
randy3k, line 93: Do we still need this for the version of lintr specified in DESCRIPTION?

renkun-ken: Now we could safely assume `lintr(text=)`. I'll remove those checking.

randy3k:
LGTM

randy3k (approved):

randy3k:
I just realized that there is no roadmap for the release of new version of `lintr`. Perhaps we should include your original trick of checking the `text` parameter for now? Otherwise we won't know how much time they need to make a new release.

Issue #283, opened 2020-06-23
On save, lintr raises "unexpected '/'" error


otherdoug:
### Issue

When saving R script files in Atom, `lintr` called via `languageserver` output reports an error `unexpected '/'` in line 1 regardless of script content.
<img width="687" alt="lintr with unexpected front-slash error" src="https://user-images.githubusercontent.com/53023343/85463199-fb45b880-b56b-11ea-8aca-3d0b81feee34.png">

Otherwise `lintr` output generally works as expected when making inline edits. Further typing without saving runs `lintr` again, and produces accurate results:
<img width="692" alt="lintr with accurate output" src="https://user-images.githubusercontent.com/53023343/85463321-20d2c200-b56c-11ea-82e3-63b3db7d1ce6.png">

### Potential cause

The `unexpected '/'` error appears to come from running `lintr::lint` on a file URI that looks like the text of an R script instead of a path to an R script:
```R
> # File URI string with newline at the end
> lintr::lint('file:///Users/myusername/local/path/to/the/file/deleteme.R\n', linters = NULL)
<text>:1:6: error: unexpected '/'
file:///Users/myusername/local/path/to/the/file/deleteme.R

> # File path
> lintr::lint('/Users/myusername/local/path/to/the/file/deleteme.R', linters = NULL)
/Users/myusername/local/path/to/the/file/deleteme.R:5:1: style: Lines should not be more than 80 characters.
tmp <- "This string is so long that it should raise a linter info message, but it won't."
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 
```

The did-save / text-sync interface appears to be origin. The interface stores the file URI in `params$textDocument` and stores what I assume is the document content in `params$text` though I have not been able to devise an example that actually stores anything there.
https://github.com/REditorSupport/languageserver/blob/61259eb9d789bdc2be7f13e4cf19ff209d5feec3/R/interfaces.R#L378-L386

When `text_document_did_save(...)` is called, `params$text` partial matches with `params$textDocument` when `params$text` is undefined. This causes `content` to populate from `stringi::stri_split_lines(text)[[1]]` by splitting the file URI as text, rather than loading the file URI from the path `stringi::stri_read_lines(path)`.
[...]

PR #291, opened 2020-06-29
Fix for #283 - "unexpected '/'" on save

otherdoug:
This pull request resolves issue #283.

It replaces `$` accessors with `[[` to prevent partial matching of `params$text` to `params$textDocument` when `text` is undefined.

Todo:
- [ ] add regression test

### DIFF
--- a/R/handlers-textsync.R
+++ b/R/handlers-textsync.R
@@ -61,9 +61,9 @@ text_document_will_save <- function(self, params) {
 #' Handler to the `textDocument/didSave` [Notification].
 #' @keywords internal
 text_document_did_save <- function(self, params) {
-    textDocument <- params$textDocument
-    text <- params$text
-    uri <- uri_escape_unicode(textDocument$uri)
+    textDocument <- params[["textDocument"]]
+    text <- params[["text"]]
+    uri <- uri_escape_unicode(textDocument[["uri"]])
     logger$info("did save:", list(uri = uri))
     path <- path_from_uri(uri)
     if (!is.null(text)) {


otherdoug:
Re: regression testing, I'm probably missing something obvious--is there a recommended way to get a properly initialized `LanguageServer` object to pass to handler functions for unit testing i.e., `self` in `text_document_did_save(self, params)`? I didn't see an obvious function in `helper-utils.R`.

Invoking `text_document_did_save` directly on a new `LanguageServer` instance raises an `attempt to apply non-function` error when attempting to call `self$workspace$documents$has(uri)` because `workspace` and `documents` are both null.

```R
server <- LanguageServer$new(host = "localhost", port = NULL)
params <- did_save_text_document_params(uri = list(uri = "file:///tmp/test.csv"), NULL)
text_document_did_save(server, params)
```

One solution would be to add mocks to the `LanguageServer` object to pass the various `has`/`set`/`get` calls, etc., but if there's an approach you all use or recommend for testing on server instances more generally, I would think that would be preferred.

renkun-ken:
Thanks for the fix!

The test cases are at `tests/testthat/test-*.R` where a language server and language client are created for testing.

This fix is minimal and obvious so I think it is okay that we skip the test cases.

If you'd like to add a test case for this, please use the functions defined in `tests/testthat/helper-utils.R` and take a look at existing test cases.

otherdoug:
Agreed it's a very minor change. If you all are OK with skipping a test case, I'll send it through.

renkun-ken:
LGTM.

@randy3k Do you think we need to use `[[` consistently in the handler functions, or we just leave as it is?

randy3k:
@renkun-ken I think that's a great idea. It will also avoid any other future related bugs. But I guess we could do the refactoring incrementally.

renkun-ken:
We could merge this first and do some refactoring later.

PR #302, opened 2020-07-06
Support startup library specified in profile

### LINKED ISSUE #301: Respect initially attached packages in .Rprofile
It seems that the initial packages to attach on session startup specified in `.Rprofile` are not respected.

For example, if user use the following `.Rprofile`:

```r
library(data.table)
```

Then the auto-completion does not include exported objects of `data.table` in the document.

renkun-ken:
Close #301

This PR respects the attached packages specified in user/system profile so that the initial auto-completion could include the package exported objects on startup.


randy3k:
Looks reasonable.

PR #305, opened 2020-07-07
Convert roxygen comments to documentation

### LINKED ISSUE #304: Roxygen example code messes up documentation
Currently, the workspace documentation is simply the uncommented text of comment above a symbol. This works if the comment is written in markdown, but it does not work well with content in roxygen `@examples` tag where R code is supplied.

For example, the following code

```r
#' Session Pool with Many R Sessions
#'
#' @examples
#' \dontrun{
#' # create 3 cached r sessions in the pool
#' pool_size <- 3
#' pool_name <- "common"
#' pool <- SessionPool$new(pool_size, pool_name)
#'
#' # check idle_size before acquiring session
#' n <- pool$get_idle_size()
#'
#' # if there are idle sessions, acquire session
#' session <- pool$acquire()
#'
#' # check session is not null and then use session
#' is.null(session)
#'
#' # use session
#' # please read R6 Class `Session` documentation
#'
#' }
#' @keywords internal
SessionPool <- R6::R6Class("SessionPool", ...)
```

its documentation is presented in hover like

image

If we could recognize the code part and use markdown code chunk to enclose it and it would certainly look better.

renkun-ken:
Close #304

This PR imports `roxygen2` and implements `convert_comment_to_documentation` where `roxygen2::parse_text()` is tried to parse comments as roxygen tags to provide structured documentation.

The information presented in hover contains all parsed roxygen tags presented in markdown format while only description is presented in completion resolve of symbol and function arguments and function signature, which is consistent with currently presented information provided by these providers.

Following is a screencast:

Kapture 2020-07-07 at 22 47 05

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,6 +30,7 @@ Imports:
     parallel,
     R6 (>= 2.4.1),
     repr (>= 1.1.0),
+    roxygen2 (>= 7.0.0),
     stringi (>= 1.1.7),
     styler (>= 1.2.0),
     tools,
--- a/R/completion.R
+++ b/R/completion.R
@@ -424,7 +424,7 @@ completion_item_resolve_reply <- function(id, workspace, params) {
             doc_string <- NULL
             if (is.character(doc)) {
                 doc_string <- doc
-            } else if (is.list(doc) && is.character(doc$description)) {
+            } else if (is.list(doc)) {
                 doc_string <- doc$description
             }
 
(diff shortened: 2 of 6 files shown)


renkun-ken:
There are several known limitations currently:

* The Rd tags (e.g. `\code{}`, `\emph{}`, `\url{}`) are not handled. [roxygen2md](https://github.com/r-lib/roxygen2md) could convert the tags but I'd rather leave them unhandled at the moment to keep the implementation clean.
* [R6 roxygen tags](https://roxygen2.r-lib.org/articles/rd.html#r6) could not non-continuous lines of roxygen comments. I don't handle such case at the moment but only takes the comment lines above the symbol definition, like other symbols.

I believe this PR is ready to merge. We could always improve the implementation if requested.

PR #309, opened 2020-07-10
Support folding ranges for comments

renkun-ken:
This PR adds support of folding ranges for comments so that consecutive lines of comments that start with the same column can be folded.

```r
# comment range 1
# comment range 1
a <- 1

# comment range 2
# comment range 2
b <- 2

# not comment range
c <- 3

f <- function(x, y) {
  # comment range 3
  # comment range 3
  x + y
}
```

The comments in the same folding ranges could be folded.

Kapture 2020-07-10 at 21 54 39


randy3k:
LGTM

PR #311, opened 2020-07-13
Improve xpath to work with cursor on token ending

renkun-ken:
Close https://github.com/Ikuyadeu/vscode-R/issues/379.

`xdoc_find_token` and `xdoc_find_enclosing_scopes` are improved to work with cursor in the end of the token, which is consistent with how other language servers deal with tokens (e.g. C++ language server works from both beginning and end).


randy3k:
Thank you for working on it.

PR #317, opened 2020-08-01
Improve folding

### LINKED ISSUE #316: Overlapping folding ranges
Currently, the folding ranges are created from all multi-line expressions and there could be overlaps in the parse data.

For example, consider the following if-else blocks:

```r
if (x > 0) {
  # hello
  # hello2
  x + 1
} else if (x < -1) {
  x - 1
} else {
  x
}
```

Collapsing a block could end up hiding the whole expression as demonstrated below:

image

It is a bit non-standard compared with the folding ranges provided by other language servers.

renkun-ken:
Close #316 

This PR make folding range provider only create folding ranges for multi-line expressions with `()`, `[]` and `{}`, and the folding range end is a line above the closing bracket so when the range is folded the closing bracket could be seen. Now the folding ranges are consistent with behavior of most other language servers.

```r
if (x > 0) {
  # hello
  # hello2
  x + 1
} else if (x < -1) {
  x - 1
} else {
  x
}
```

Before:

image

After:

image

And multi-line expressions not enclosed in a pair of brackets are no longer foldable. For example, a pipeline expression could no longer be folded as it is not enclosed in brackets.

```r
kable(dt, format = "html", caption = "Demo Table") %>%
  kable_styling(bootstrap_options = "striped",
    full_width = FALSE) %>%
  add_header_above(c(" ", "Group 1" = 2, "Group 2[note]" = 2)) %>%
  add_footnote(c("table footnote"))
```

### DIFF
--- a/R/folding.R
+++ b/R/folding.R
@@ -5,19 +5,27 @@ FoldingRangeKind <- list(
 )
 
 get_expr_folding_ranges <- function(xdoc) {
-    exprs <- xml_find_all(xdoc, "//expr[@line2 > @line1]")
+    exprs <- xml_find_all(xdoc, "//expr[@line1 < @line2 and
+        (OP-LEFT-PAREN | OP-LEFT-BRACKET | OP-LEFT-BRACE)/@line1 <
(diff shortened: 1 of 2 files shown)


randy3k:
Please just merge it once you feel good about it. :)

PR #323, opened 2020-08-19
Only provide imported completions without package

### LINKED ISSUE #322: Imported objects appear in completion after ::
#248 introduces the handling of `NAMESPACE` file so that imported objects could appear in the completion list.

However, if we are working on a package project, then those imported objects will always appear after `::` or `:::` no matter if the package name contains those objects.

Following are some images to demonstrate this bug while working on languageserver which imports `R6Class` from `R6`:

image

image

image

renkun-ken:
Close #322

### DIFF
--- a/R/completion.R
+++ b/R/completion.R
@@ -265,11 +265,10 @@ workspace_completion <- function(workspace, token,
         }
     }
 
-    imported_object <- imported_object_completion(workspace, token, snippet_support)
-
-    completions <- c(
-        completions,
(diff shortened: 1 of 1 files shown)


randy3k:
Thank you for catching it.

randy3k:
I think it's about time to make a new release.

renkun-ken:
>I think it's about time to make a new release.

Yes, definitely.

Issue #158, opened 2020-01-01
Provide completion for list-like objects in package


renkun-ken:
We may provide completion for list-like objects in package such as `.Machine`, `.Platform`, and objects defined in packages such as R6 classes since their elements are known.

In contrast with https://github.com/Ikuyadeu/vscode-R/pull/165 which provides completion for elements in list-like objects in a live R session, languageserver may do similar for list-like objects in packages.

qinwf:
With typescript style `type definition file` will help simplify the work, aka `typedef` for R object.

R6 classes are already typed. They will be easier to process.

renkun-ken:
Sounds interesting, would you like to elaborate?

qinwf:
https://devblogs.microsoft.com/python/announcing-pylance-fast-feature-rich-language-support-for-python-in-visual-studio-code/

Python introduced  https://docs.python.org/3/library/typing.html

and pylance is using https://github.com/microsoft/pyright for the information, which is similar to typescript and javascript.

renkun-ken:
@qinwf Just noticed Pylance yesterday. Looks quite promising. Do you have any idea how we could benefit from its design or implementation?

qinwf:
Pylance is based on type hints featured built into Python 3.5+. 

There is no typing library for R now. A gradual type system is required to enable features like it. 

To build a gradual type system, there will be a frontend to do type checking for R code without running the code.

Issue #75, opened 2019-09-27
Auto-completion for R6 object


hongyuanjia:
Thanks for this amazing package. It works brilliant for Vim on Linux. It seems like that auto-completion on R6 object methods is not supported. Is there any plan to add support for R6 object? Thanks!

randy3k:
It is not trivial. The easy part is to parse an R6 class and inspect its methods. The more tricky part is to deduce if a variable is an R6 object or not. It’s not always as easy as 
```
var <- MyClass$new()
```
In particular, a variable could be assigned to any thing dynamically. With only lexical analysis, it’s not 100% guaranteed that the inspected result is correct. An simple example could be
```
var <- if (condition) MyClass$new() else AnotherClass$new()
``` 
Not to mention that the variable could be a result of a complicated chain of actions.

One resolution that I am thinking is manual explicit annotations via roxygen

```
@param x MyClass
@return AnotherClass
foo <- function(x) {
      AnotherClass$new(x)
}
```

Of course, this method also comes with its problems.

hongyuanjia:
Thanks for the explanation.

Sorry I did not know the internal implementation of auto-completion. I am not sure if current implementation is totally different from what [RStudio does](https://github.com/rstudio/rstudio/blob/master/src/cpp/session/modules/SessionRCompletions.R). Probably I am underestimate the difficulty.

randy3k:
What you are pointing to is related to console auto completion. In a console, we know exactly if an object is R6 or not. However, we are working with editors and we don’t actually run the code. All the analysis are done lexically which makes the job difficult.

hongyuanjia:
Oh I see where the difficulty comes from. This is probably a restriction when working on lsp. Anyway, still want to thank you for this amazing work and hope this feature can be implemented in the future.

hongyuanjia:
> One resolution that I am thinking is manual explicit annotations via roxygen
> 
> ```
> @param x MyClass
> @return AnotherClass
> foo <- function(x) {
>       AnotherClass$new(x)
> }
> ```

Just see that roxygen2 now supports more intuitive documentation via https://github.com/r-lib/roxygen2/pull/922. Will it help to implement R6 autocomplete?

randy3k:
Yes, every extra bit of information will help.

renkun-ken:
[...]

PR #324, opened 2020-08-21
Add token completion

renkun-ken:
It is hard to provide accurate completion based on static code analysis in the following cases:

* Completion for R6 class members (#75)
* Completion for list or data.frame elements (#158)
* Completion for symbols in a pipeline (https://github.com/Ikuyadeu/vscode-R/issues/323)

This PR is an attempt to partially address these cases by providing completions from symbols that appear in the current document so that the symbols are recognized as texts that could be repeatedly used in the document, as demonstrated in the following images.

image

image

Since the token completion items have the lowest ranking, they won't interrupt normal completions but provide some help when the normal code analysis does not provide anything.

### DIFF
--- a/R/completion.R
+++ b/R/completion.R
@@ -373,7 +373,7 @@ completion_reply <- function(id, uri, workspace, document, point, capabilities) 
     token <- token_result$token
     package <- token_result$package
 
-    completions <- token_completion(uri, workspace, token)
+    completions <- list()
 
     if (nzchar(full_token)) {
         if (is.null(package)) {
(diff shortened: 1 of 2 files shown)


renkun-ken:
I've been trying this for several days in my work and the completions works well for me as a good supplement for cases where static analysis does not yield useful results.

randy3k:
Thank you for testing it. It seems to break some of the tests tough.

renkun-ken:
The Windows tests seem not always stuck at  `Run r-lib/actions/setup-r@v1`, not sure why.

PR #338, opened 2020-09-19
Implement referencesProvider

### LINKED ISSUE #336: Implement referencesProvider
>The references request is sent from the client to the server to resolve project-wide references for the symbol denoted by the given text document position.

The [referencesProvider](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_references) could be implemented to have the following behavior:

1. If the workspace is an R package, then find references in all R documents; otherwise, find references in open documents.
2. Get the token at the requested position.
3. Use the approach of definition provider to find definition of the token.
4. Find all symbols in the documents and go-to-definition of each symbol.
5. All symbols that go to the same definition will be included as a reference of the original token.

renkun-ken:
Close #336 

This PR implements a simple [referencesProvider](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_references) based on the current definitionProvider.

1. Find the definition of the requested token.
2. Find all symbols with the same text of the requested token in all documents in workspace (`R/*.R` for package or otherwise open R documents in non-package workspace)
3. Find the definition of each symbol in each document.
4. Regard the symbol as a reference if its definition is identical to that of the requested token.

This PR also provides test cases for the referencesProvider (also based on test cases of the definitionProvider).

Kapture 2020-09-19 at 20 39 47

### DIFF
--- a/R/capabilities.R
+++ b/R/capabilities.R
@@ -50,7 +50,7 @@ ServerCapabilities <- list(
     # typeDefinitionProvider = FALSE,
     # implementationProvider = FALSE,
     definitionProvider = TRUE,
-    # referencesProvider = FALSE
+    referencesProvider = TRUE,
     documentHighlightProvider = TRUE,
     documentSymbolProvider = TRUE,
     workspaceSymbolProvider = TRUE,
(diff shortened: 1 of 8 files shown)


randy3k:
it seems that the tests have failed the coverage tests. I will take a look later today.

renkun-ken:
There seems to be some difference between the coverage runner and build runners.

randy3k:
LGTM. Merge it at your will.

randy3k (approved):

PR #339, opened 2020-09-19
Implement renameProvider

### LINKED ISSUE #337: Implement renameProvider
> The rename request is sent from the client to the server to ask the server to compute a workspace change so that the client can perform a workspace-wide rename of a symbol.

The [rename provider](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_rename) could be implemented via the same approach of #336.

renkun-ken:
Close #337 

This PR implements [prepareRename](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_prepareRename) and [renameProvider](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_rename) based on the referenceProvider implemented by #338 by transforming the references into a [WorkspaceEdit](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspaceEdit).

Kapture 2020-09-19 at 21 55 06

### DIFF
--- a/R/handlers-general.R
+++ b/R/handlers-general.R
@@ -17,10 +17,12 @@ on_initialize <- function(self, id, params) {
     self$workspace <- Workspace$new(self$rootPath)
     self$initializationOptions <- params$initializationOptions
     self$ClientCapabilities <- params$capabilities
-    ServerCapabilities <- merge_list(
-        ServerCapabilities,
+    server_capabilities <- update_server_capabilities(
+        ServerCapabilities, self$ClientCapabilities)
+    server_capabilities <- merge_list(
+        server_capabilities,
         getOption("languageserver.server_capabilities"))
-    self$deliver(Response$new(id = id, result = list(capabilities = ServerCapabilities)))
+    self$deliver(Response$new(id = id, result = list(capabilities = server_capabilities)))
 }
 
 #' `initialized` handler
--- a/R/handlers-general.R
+++ b/R/handlers-general.R
@@ -22,6 +22,7 @@ on_initialize <- function(self, id, params) {
     server_capabilities <- merge_list(
         server_capabilities,
         getOption("languageserver.server_capabilities"))
+    self$ServerCapabilities <- server_capabilities
     self$deliver(Response$new(id = id, result = list(capabilities = server_capabilities)))
 }
 
--- a/tests/testthat/test-rename.R
+++ b/tests/testthat/test-rename.R
@@ -1,0 +1,271 @@
+context("Test Rename")
+
+test_that("Rename works for functions in files", {
+    skip_on_cran()
+    client <- language_client()
+
+    withr::local_tempfile(c("defn_file", "defn2_file", "query_file"), fileext = ".R")
+    writeLines(c("my_fn <- function(x) {", "  x + 1", "}"), defn_file)
+    writeLines(c("my_fn"), query_file)
+
+    client %>% did_save(defn_file)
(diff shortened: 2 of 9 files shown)


randy3k:
Sorry for the slow review. Will try to find some time to do it tomorrow.

renkun-ken:
Feel free to take your time!

We might review and merge in the following order due to dependencies:

#338, #339

tests/testthat/test-rename.R
```diff
@@ -0,0 +1,298 @@
...
+
+    result <- client %>% respond_prepare_rename(single_file, c(1, 0))
+    expect_equal(result$start, list(line = 1, character = 0))
+    expect_equal(result$end, list(line = 1, character = 5))
+
+    result <- client %>% respond_prepare_rename(single_file, c(0, 0))
+    expect_equal(result$start, list(line = 0, character = 0))
+    expect_equal(result$end, list(line = 0, character = 5))
+
+    result <- client %>% respond_prepare_rename(single_file, c(2, 0))
```
renkun-ken, line 292: The [prepareRename](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_prepareRename) says that the result should be `null` and an error code/message should be provided if the given position cannot be renamed.

However, this call to `respond_prepare_rename` returns a response with `error` and then it does not call the `callback` function to ensure the request is finished in the testing. Should we alter the testing utility so that it allows callback to be called when an error is returned.

renkun-ken: I added a `allow_error` argument to `respond` so that error is allowed if it is expected.

R/handlers-general.R
```diff
@@ -17,10 +17,12 @@ on_initialize <- function(self, id, params) {
     self$workspace <- Workspace$new(self$rootPath)
     self$initializationOptions <- params$initializationOptions
     self$ClientCapabilities <- params$capabilities
-    ServerCapabilities <- merge_list(
-        ServerCapabilities,
+    server_capabilities <- update_server_capabilities(
+        ServerCapabilities, self$ClientCapabilities)
+    server_capabilities <- merge_list(
```
randy3k, line 22: Do we want to keep a copy of it in `self`?

renkun-ken: It makes sense. I added `ServerCapabilities` to `LanguageServer` too to store it.

renkun-ken: @randy3k With 33f6c09, does it look good now?

renkun-ken:
@randy3k Is there anything you think should be improved?

randy3k:
LGTM. Sorry for missing it.

randy3k (approved):

randy3k:
By the way, how's its performance for a large project?

renkun-ken:
renameProvider relies on referencesProivder which again relies on definitionProvider. The performance of renameProvider is almost equal to that of the referencesProvider.

I tried using the `Find All References` in projects languageserver, lintr, data.table, ggplot2, and knitr, and the performance looks good in most cases. In fact, its performance could be poor if we try to find a token which occurs too many times (e.g. more than 1000+) in the projects.

randy3k:
> In fact, its performance could be poor if we try to find a token which occurs too many times (e.g. more than 1000+) in the projects.

That's what I was referring to. But anyway, such of use case should not be common.

grepinsight:
👍 👍  Thank you for this feature

ManuelHentschel:
Great feature, thanks a lot! 👍

PR #342, opened 2020-09-22
Fix checking symbol on rhs of assignment expression

### LINKED ISSUE #341: Check if symbol on rhs of assignment in definition
Consider the following code:

```r
test <- 1
test <- test + 1
```

The definition should be following:

1. `test` in line1: line 1
2. `test` in line2 before `<-`: line 2
3. `test` in line2 after `<-`: line 1

The current implementation of definition violates 3 as it will ignore whether the token is on the right side of the assignment expression.

renkun-ken:
Close #341 

This PR improves the XPaths of definition and hover so that the XPaths ensure that the requested token is not included in the value expression of the assignment, i.e. RHS of left assignment and equal assignment and LHS of right assignment.

Then the definition and hover providers will regard line 1 as the definition of token `test` in line 2 on RHS of the assignment:

```r
test <- 1
test <- test + 1
```


randy3k (approved):
LGTM. Actually, the XPATH syntax is too complex for me to review 😓

PR #347, opened 2020-10-01
Do not load sysdata.rda

jozefhajnala:
In the R directory, some packages have the `sysdata.rda` file (see https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Package-subdirectories for reference), which tries to get loaded on_initialized. This PR proposes to skip that file from `load_all()`.

### DIFF
--- a/R/workspace.R
+++ b/R/workspace.R
@@ -226,7 +226,7 @@ Workspace <- R6::R6Class("Workspace",
 
         load_all = function(langserver) {
             source_dir <- file.path(self$root, "R")
-            files <- list.files(source_dir)
+            files <- list.files(source_dir, pattern = "\\.r$", ignore.case = TRUE)
             for (f in files) {
                 logger$info("load ", f)
                 path <- file.path(source_dir, f)


randy3k:
a better solution may be to load only files end with `.R` or `.r`.

jozefhajnala:
> a better solution may be to load only files end with `.R` or `.r`.

I wanted to err on the side of conservativism, as there could also be extensions `.S`, `.q`, or `.s` for source files and `.in` files for configure scripts, but of course if you are happy with just loading `.R` or `.r` I can update the PR.

jozefhajnala:
I updated the PR, if you would like to re-review.

renkun-ken (approved):
LGTM

renkun-ken:
@randy3k Do we need to resolve the CI issue before merging?

randy3k:
This one looks good to me.

randy3k:
> there could also be extensions `.S`, `.q`, or `.s` for source files.

I don't think it makes sense to support all the ancient file types. Anyway, thank you for your contribution.

PR #350, opened 2020-10-16
Update coc help in README

she3o:
Updates README because the original didn't show that coc could be configured entirely from `coc-settings.json`.


randy3k:
thanks.

PR #351, opened 2020-10-28
Fix tests

randy3k:
- update syntax for withr::local_tempfile

- do not expect returns when shutting down server

- need to retry in various places to fix the coverage tests because tests running with covr are slow


renkun-ken:
Nice work to make the CI runners succeed again!

renkun-ken (approved):
LGTM

PR #363, opened 2021-01-08
use expect_setequal() may fit better

shrektan:
I think using `expect_setequal()` may be better for these testing cases.


randy3k:
Agree.

renkun-ken (approved):
LGTM

PR #364, opened 2021-01-08
Recognize symbols created by delayedAssign / assign / makeActiveBinding

### LINKED ISSUE #362: [FR] Display promise / active-binding objects in the OUTLINE
Considering the following test file: 

**test.R**

```r
x <- 1
y <- "a"
z <- function() {
  'hello'
}
delayedAssign("d", 1)
delayedAssign("err", stop("err"))
makeActiveBinding("a1", function() 1, environment())
```

Currently, only `x`, `y` and `z` are displayed in the OUTLINE but `d`, `err` and `a1` don't.

I hope we can have them in the OUTLINE, too.

image

shrektan:
Closes #362 

## A simple test file

```r
x <- 1
y <- "a"
z <- function() {
  'hello'
}
delayedAssign("d1", 1)
delayedAssign(value = function() 2, x = "d2")
base::delayedAssign(value = "3", "d3")
var <- "d4"
delayedAssign(var, 4)
delayedAssign(("d5"), var)

makeActiveBinding("a1", function() 1, environment())
makeActiveBinding(function() "2", sym = "a2")
base::makeActiveBinding(fun = function() stop("3"), sym = "a3")
makeActiveBinding(("a4"), function() 1, environment())

assign(value = "1", "assign1")
```

## Outline can display the promise / active binding symbols now

image

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,14 +1,17 @@
+# languageserver 0.3.10
+
+- Recongize promise and active binding symbols (#362)
+
 # languageserver 0.3.9
 
 - skip tests on solaris 
-- Recongize promise and active binding symbols (#362)
 
 # languageserver 0.3.8
 
 - When closing a file, "Problems" should be removed (#348)
 - Implement renameProvider (#337)
 - Hover on symbol in a function with functional argument causes parse error (#345)
-Hover on non-function symbol in other document should show definition and - documentation (#343)
+- Hover on non-function symbol in other document should show definition and - documentation (#343)
 - Check if symbol on rhs of assignment in definition (#341)
 - Implement referencesProvider (#336)
 - Add comment of notice above temp code of definition (#353)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # languageserver 0.3.9
 
 - skip tests on solaris 
+- Recongize promise and active binding symbols (#362)
 
 # languageserver 0.3.8
 
--- a/R/document.R
+++ b/R/document.R
@@ -218,11 +218,9 @@ parse_expr <- function(content, expr, env, level = 0L, srcref = attr(expr, "srcr
             Recall(content, e[[3L]], env, level + 1L, cur_srcref)
         } else if (f == "repeat") {
             Recall(content, e[[2L]], env, level + 1L, cur_srcref)
-        } else if (
-            (f %in% c("<-", "=") && length(e) == 3L && is.symbol(e[[2L]])) ||
-            f %in% c("delayedAssign", "makeActiveBinding")
-        ) {
+        } else if (f %in% c("<-", "=", "delayedAssign", "makeActiveBinding")) {
             if (f %in% c("<-", "=")) {
+                if (length(e) != 3L || !is.symbol(e[[2L]])) next
                 symbol <- as.character(e[[2L]])
                 value <- e[[3L]]
             } else if (f == "delayedAssign") {
--- a/R/document.R
+++ b/R/document.R
@@ -224,7 +224,13 @@ parse_expr <- function(content, expr, env, level = 0L, srcref = attr(expr, "srcr
             # if set, we should compare to a vector of known "top-level" candidates
             is_top_level <- function(arg_env, ...) {
                 if (is.null(arg_env)) return(TRUE)
-                top_level_envs <- substitute(list(...))[-1L]
+                default <- list(
+                    quote(parent.frame(1)), quote(parent.frame(1L)),
+                    quote(environment()),
+                    quote(.GlobalEnv), quote(globalenv())
+                )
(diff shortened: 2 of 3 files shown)


R/document.R
```diff
@@ -204,9 +218,23 @@ parse_expr <- function(content, expr, env, level = 0L, srcref = attr(expr, "srcr
             Recall(content, e[[3L]], env, level + 1L, cur_srcref)
         } else if (f == "repeat") {
             Recall(content, e[[2L]], env, level + 1L, cur_srcref)
-        } else if (f %in% c("<-", "=") && length(e) == 3L && is.symbol(e[[2L]])) {
-            symbol <- as.character(e[[2L]])
-            value <- e[[3L]]
+        } else if (
+            (f %in% c("<-", "=") && length(e) == 3L && is.symbol(e[[2L]])) |
```
renkun-ken, line 222: You might want to use `||` instead of `|`?

shrektan: Correct!

NEWS.md
```diff
@@ -1,6 +1,7 @@
 # languageserver 0.3.9
 
 - skip tests on solaris 
+- Recongize promise and active binding symbols (#362)
```
randy3k, line 4: this shoul goes to 0.3.10 :-)

renkun-ken:
It does not seem to work with the following:

```r
makeActiveBinding(("a"), function() 1, environment())
```

We could simply ignore these calls with the name being not simply a string literal.

renkun-ken:
I just recall the reason we didn't detect these assign functions. Consider the following examples:

```r
e <- new.env()
assign("var", 1, e)
assign("var", 2, globalenv())
assign("var", 3, .GlobalEnv)
```

All these assign functions have an argument to specific in which environment the assignment should be done. The document outline only lists the global symbols in the document. However, these assignments may be done in any environment. Should we detect the `pos` or `envir` argument?

shrektan:
Yes, I'm aware of this issue. My personal points are:

1. The outlined symbols are not necessarily "global" - as the script may be used in an R package or be `source()` with `local=TRUE`. So maybe we should call them as  "top-level" symbols.
2. Except `assign()`, I guess in majority cases, users use  `delayedAssign()` or `makeActiveBinding()` on the top level to make "top-level" variables. 
3. Greedy may be a good choice for the document outline purpose.

Simple improvements for this, in my opinion, could be detecting the `pos` / `envir` / `assign.env` argument and only mark the symbols as top-level, when these arguments:

- are not provided, or
- are provided and equal to default (`-1` / `parent.frame(1)`) or `.GlobalEnv`, etc...

However, I'm hesitate to do it as it complicates the code.

So **my suggestion** is that we may choose one of the two strategy:

1. **Greedy strategy**: Ignore the possibility that those symbols may be assigned in a different context, for the time being.  We can improve this when users complain that this causes trouble.
2. **Detecting those assigning environment arguments but using a simple strategy**. That is we may assume that *if those args are provided, it means users want to assign the symbols under a different environment* - so we don't display them in the outline. In short, **we ignore those symbols whose assigning env arguments (e.g., `pos` or `env`) are provided**.

What do you think?

renkun-ken:
>Detecting those assigning environment arguments but using a simple strategy. That is we may assume that if those args are provided, it means users want to assign the symbols to a different environment - so we don't display them in the outline. In short, we ignore those symbols whose assigning env arguments (e.g., pos or env) are provided.

I think this is good enough for most cases.

shrektan:
As `makeActiveBinding()` doesn't have a default `env`, I need to check the `env` agains a list of possible "top-level" candidates.  Thus, the checks for `assign()` and `delayedAssign()` are added in the meantime.

R/document.R
```diff
@@ -204,9 +218,46 @@ parse_expr <- function(content, expr, env, level = 0L, srcref = attr(expr, "srcr
...
+                if (!is_top_level(call$pos, -1L, -1)) next # -1 is the default
+                if (!is_top_level(call$envir)) next
+                symbol <- call$x
+                value <- call$value
+            } else if (f == "makeActiveBinding") {
+                call <- match.call(base::makeActiveBinding, as.call(e))
+                if (!is.character(call$sym)) next
+                if (!is_top_level(call$env)) next
+                symbol <- call$sym
+                value <- call$fun
```
renkun-ken, line 258: The active bindings should not be treated as a function or the completion and function signature will also apply to them but they should not.

Consider

```r
makeActiveBinding("act", function() 1, environment())
```

The completion item shows that it is a function and it will be completed as `act()`, and function signature like the following is present:

image

I guess we should alter `value` so that it is not a function and then it will be treated as a typical variable.

renkun-ken: Fixed via d15771e

renkun-ken (approved):
LGTM

PR #369, opened 2021-01-26
Add '(' to completion trigger characters

renkun-ken:
I think it makes sense to add `(` as a completion trigger character when user types a function call it will show the arguments in completion and signature help at the same time.


randy3k:
I think it may be a good default too.

randy3k (approved):

PR #370, opened 2021-01-26
Preserve argument order in argument completion

renkun-ken:
I think it looks better if the order of argument completion items is consistent with the function formals.


randy3k:
LGTM

randy3k (approved):

randy3k:
Perhaps adding a test case for this?

renkun-ken:
Not sure why build checks passed but coverage check failed?

PR #372, opened 2021-01-27
Improve token completion

renkun-ken:
This PR improves the token completion with the following changes:

1. Since #369 allows `(` to trigger completion to in accompany with https://github.com/Ikuyadeu/vscode-R/pull/530, the current token_completion may produce too many symbols once `(` triggers completion. This PR makes token_completion only show symbols after `$` and `SYMBOL_SUB` symbols like `foo` in `mutate(foo = bar)` when trigger `token` is empty (e.g. `foo$` and `fun(`), which are the most frequent ways to introduce new data variables.
2. If `token` is not empty, then all symbols from the document that start with `token` are searched, which is the original method.

This makes it possible to work with the following examples:

```r
library(data.table)
library(dplyr)
library(tibble)

tbl1 <- tibble(
  Name = c("Product1", "Product2", "Product3"),
  Quality = c("Good", "Good", "Best"),
  Score = c(8, 9, 10),
  Type = c("B", "B", "A")
)

tbl1 %>%
  mutate(Score_mean = mean(Score)) %>%
  select(|

x <- iris
x %>%
  mutate(score1 = Petal.Length + Petal.Width, score2 = Petal.Length * Petal.Width) %>%
  filter(|


df <- list(
  test1 = 1,
  test2 = 2
)
df$test3 <- 3
df$te|


dt <- data.table(var1 = 1:10, var2 = rnorm(10))
dt[, var3 := var1 + var2]
dt[, var4 := var|]
```

where `|` is the cursor and the completion should work as expected to complete the tokens from the document.

### DIFF
--- a/R/completion.R
+++ b/R/completion.R
@@ -404,15 +404,33 @@ scope_completion <- function(uri, workspace, token, point, snippet_support = NUL
     completions
 }
 
-token_completion <- function(uri, workspace, token, exclude = NULL) {
+token_completion <- function(uri, workspace, token, point, exclude = NULL) {
     xdoc <- workspace$get_parse_data(uri)$xml_doc
     if (is.null(xdoc)) {
         return(list())
     }
 
-    symbols <- unique(xml_text(xml_find_all(xdoc,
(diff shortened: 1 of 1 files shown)


randy3k (approved):
Perfect.

PR #373, opened 2021-01-29
Implement Call Hierarchy provider

### LINKED ISSUE #361: Support Call Hierarchy
In the current [Language Server Protocol Specification - 3.16](https://microsoft.github.io/language-server-protocol/specifications/specification-current/), the call hierarchy supported is added. As per requested by https://github.com/Ikuyadeu/vscode-R/issues/509, we could consider adding support for this in languageserver.

renkun-ken:
Close #361 
Close https://github.com/Ikuyadeu/vscode-R/issues/509

This PR is an initial implementation of the Call Hierarchy features.

Kapture 2021-01-31 at 13 54 12

- [x] [`textDocument/prepareCallHierarchy`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_prepareCallHierarchy)
- [x] [`callHierarchy/incomingCalls`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#callHierarchy_incomingCalls)
- [x] [`callHierarchy/outgoingCalls`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#callHierarchy_outgoingCalls)
- [x] `CallHierarchyItem` details
- [x] Handle calls from functions not defined in workspace (e.g. from list element or class methods)
- [x] Test cases


R/call_hierarchy.R
```diff
@@ -0,0 +1,228 @@
...
+          line1 = line1, col1 = col1, line2 = line2, col2 = col2,
+          token_quote = token_quote
+        )
+      )
+
+      if (length(symbols) == 0) {
+        next
+      }
+
+      defn_hash <- digest::digest(list(uri = doc_uri, definition = defn))
```
randy3k, line 84: you might also consider `collections::dict` which supports hashing arbitrary R objects.

renkun-ken: I considered about using `collections::dict` at first. Do you know a similar way using it with reference semantics like `env[[key]]$foo <- c(env[[key]]$foo, bar)` so that we don't need to re-map every element in the end to get the LSP result?

randy3k: nested dict?

```
d <- collections::dict()
d$set("foo", collections::dict())

foo <- d$get("foo")
foo$set(c(foo, bar))
```

randy3k: Actually, I don't quite understand what you mean by "remap every element".

renkun-ken: I switch to `collections::dict()`. Using `uenv` still creates new lists with shallow copy anyway.

renkun-ken:
Any other suggestions?

randy3k (approved):
LGTM. Thank you for your work.

PR #374, opened 2021-01-29
Remove '(' from completion trigger chars

renkun-ken:
Having experienced the #369, using `(` as a completion trigger character is in most cases too eager: once the completion is triggered, the new typed character won't trigger the computation of completions in a while. In practice, this will usually disable many other completion items or leave no change for them to show up.

Therefore, I'd like to revert that change by removing `(` from completion trigger characters.


randy3k:
No problem. :)

PR #382, opened 2021-02-19
Load documents referenced in source() calls

### LINKED ISSUE #20: Extend completion with source by recursive parsing
It is tricky to provide completion without attaching to the R session being used by user. One simplified approach that I believe can serve most of the purposes is parse the currently editing document and find top-level `source` functions, and parse those files recursively. 

For example, the following code is from several R scripts:

`main.R`:

```r
library(data.table)
library(glmnet)
source("my_plot.R")
source("data/utils.R")
```

`my_plot.R`:

```r
library(ggplot2)
```

`data/utils.R`:

```r
library(DBI)
requireNamespace("RMariaDB")
```

When `main.R` is being edited, those `library` expressions are captured. If those `source` expressions are captured and `my_plot.R` and `data/utils.R` are analyzed, where `library` and `source` expressions in these files are also captured, the completion will be much better and no less accurate, especially when #19 is implemented.

Also those top-level `library()`, `require()`, `loadNamespace()`, `requireNamespace()` calls can be used to provide diagnostics when required package is not installed.

renkun-ken:
Closes #20 

This PR detects top-level `source()` calls and load documents referenced in such calls.

For example,

`sources.R`:

```r
source("source1.R")
```

`source1.R`:

```r
source("source2.R")

f <- function(x) {
  x + 1
}
```

`source2.R`:

```r
g <- function(x) {
  x + 1
}
```

Opening `sources.R` will also load `source1.R` and `source2.R`.

### DIFF
--- a/R/document.R
+++ b/R/document.R
@@ -293,6 +293,11 @@ parse_expr <- function(content, expr, env, level = 0L, srcref = attr(expr, "srcr
             if (!(pkg %in% env$packages)) {
                 env$packages <- c(env$packages, pkg)
             }
+        } else if (f %in% "source") {
+            call <- match.call(base::source, as.call(e))
+            if (!is.character(call$file)) next
+            if (!is.null(call$local) && !identical(call$local, FALSE)) next
+            env$sources <- c(env$sources, call$file)
         }
     }
     env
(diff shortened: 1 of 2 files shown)


randy3k:
Didn't check the code yet. Does it handle
```r
p <- "path"
source(p)
```

renkun-ken:
Closing as the design is still problematic.

PR #389, opened 2021-03-13
Options completion

renkun-ken:
This PR adds arg completion for `options()` calls so that it is easier to complete options.

Note that if user code calls e.g. `library(data.table)`, and the new options defined in `data.table` will also appear in the options arg completion since the package is also loaded by languageserver.

### DIFF
--- a/R/completion.R
+++ b/R/completion.R
@@ -114,7 +114,7 @@ arg_completion <- function(uri, workspace, point, token, funct, package = NULL, 
         args <- names(workspace$get_formals(funct, package, exported_only = exported_only))
 
         if (package == "base" && funct == "options") {
-            args <- c(args, names(options()))
+            args <- c(args, names(.Options))
         }
 
         if (is.character(args)) {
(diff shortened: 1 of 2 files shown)


ManuelHentschel:
Works nicely, thanks!

Would you consider listing the [options used in base R](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/options) that are not set initially? (Might be useful, but would also require keeping the list up to date and hence introduce a source of inconsistencies)

Also I saw that in the help for `options()` that using `.Options` is apparently faster than calling `options()`, in case this (probably rather small) performance difference is of any concern here.

renkun-ken:
> Would you consider listing the options used in base R that are not set initially? (Might be useful, but would also require keeping the list up to date and hence introduce a source of inconsistencies)

Do we have a programmatic way to do this like parsing R source files, find all `getOption()` calls and extract those names? Maybe it is worth another PR to do this?

> Also I saw that in the help for options() that using .Options is apparently faster than calling options(), in case this (probably rather small) performance difference is of any concern here.

Yes, it is definitely faster. Thanks!

ManuelHentschel:
> Do we have a programmatic way to do this like parsing R source files, find all `getOption()` calls and extract those names? Maybe it is worth another PR to do this?

That sounds possible, but would probably warrant a separate PR as you suggested

ManuelHentschel (approved):

PR #393, opened 2021-03-18
Limit link file size

### LINKED ISSUE #391: Document link should work with limited file extensions or file size
Currently, the document link provider will capture all existing file paths in the document to provide links. Some file paths are binary data files (e.g. `*.fst`) rather than source files (`*.R`) or text-data files (e.g. `*.csv`). Accidentally following the link of data binary data file is not useful (the editor cannot handle the file type) or sometimes makes the editor freeze (via remote-ssh).

I think we should use an option to provide links to a limited collection of file extensions. By default, it might include R, csv, log, txt, etc.

renkun-ken:
Closes #391 

The main reason not to use file extension is that there are too many text file extensions so that it might not be a good idea. To keep user from following a link to a large file that might freeze the editor, we only need to avoid providing links to large files.

This PR checks file size at `documentLink/resolve` so that large files won't be provided links. By default, the file size limit is 16MB. User could customize it via `options(languageserver.link_file_size_limit = bytes)`.

The reason to check file size at resolve stage is that we could let the user know that the file exists via editor text decoration (e.g. text underlined in vscode).

### DIFF
--- a/R/capabilities.R
+++ b/R/capabilities.R
@@ -35,7 +35,7 @@ DocumentOnTypeFormattingOptions <- list(
 )
 
 DocumentLinkOptions <- list(
-    resolveProvider = FALSE
+    resolveProvider = TRUE
 )
 
 RenameOptions <- list(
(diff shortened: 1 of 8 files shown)


randy3k:
LGTM

PR #404, opened 2021-03-30
Use styler base_indention

### LINKED ISSUE #400: Use `styler::style_text(base_indention=)` for `styler > 1.4.0`
`base_indention` is supported via https://github.com/r-lib/styler/pull/649 since styler 1.4.0. We may consider using this instead of hacking into the stylers for `styler >= 1.4.0` in the formatting providers.

renkun-ken:
Closes #400.

This PR uses the latest styler `base_indention` and removes the old hacking way to add indention.

I think we could wait until https://github.com/r-lib/styler/pull/764 is merged upstream so that this PR could be merged too.

We'll update the minimum version of styler in `DESCRIPTION` if a new version (probably v1.4.1) is released to CRAN.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -46,6 +46,8 @@ Suggests:
     testthat (>= 2.1.0),
     withr (>= 2.3.0),
     rmarkdown (>= 2.0)
+Remotes:
+    r-lib/styler#764
 ByteCompile: yes
 Encoding: UTF-8
 LazyData: true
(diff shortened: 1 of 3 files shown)


randy3k:
Perhaps we could specify the following item that `devtools::install` supports
```
Remotes: r-lib/styler#764
```

see https://cran.r-project.org/web/packages/devtools/vignettes/dependencies.html for details

renkun-ken:
Do we also include this PR in the upcoming release? The macOS and Windows build is not ready due to the binary builds are still on the way.

randy3k:
Let's wait for the binaries for styler then.

renkun-ken:
The binary builds on all platforms are ready and the CI passes now.

renkun-ken:
@randy3k Looks like we could do a new release then.

PR #405, opened 2021-03-30
Provide case-insensitive completion

### LINKED ISSUE #401: Case-insensitive and fuzzy match in completion
Currently, we use `startsWith()` to filter completion items. It is fast, but it is case-sensitive and does not allow imprecise matches with the first few characters.

Maybe we should allow some fuzziness and case-insensitivity?

renkun-ken:
Closes #401.

This PR uses `grep(ignore.case=TRUE)` to provide case-insensitive and partial matching completion items to replace the current case-sensitive, exact-matching-from-start filtering.

* [x] Change matching
* [x] Update test cases

### DIFF
--- a/R/completion.R
+++ b/R/completion.R
@@ -50,7 +50,7 @@ constants <- c("TRUE", "FALSE", "NULL",
 #' Complete language constants
 #' @keywords internal
 constant_completion <- function(token) {
-    consts <- constants[startsWith(constants, token)]
+    consts <- constants[match_with(constants, token)]
     completions <- lapply(consts, function(const) {
         list(label = const,
             kind = CompletionItemKind$Constant,
@@ -64,7 +64,7 @@ constant_completion <- function(token) {
 #' @keywords internal
 package_completion <- function(token) {
     installed_packages <- .packages(all.available = TRUE)
-    token_packages <- installed_packages[startsWith(installed_packages, token)]
+    token_packages <- installed_packages[match_with(installed_packages, token)]
     completions <- lapply(token_packages, function(package) {
         list(label = package,
             kind = CompletionItemKind$Module,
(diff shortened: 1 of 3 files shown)


randy3k:
It may be desired for the exact matches to show first. Though it may need some moderate refactoring.

renkun-ken:
> It may be desired for the exact matches to show first. Though it may need some moderate refactoring.

I don't know how other editors handle the order, vscode already does it by choosing the exact match first.

renkun-ken:
image

The matches are case-insensitive. The first few are matches from the start, then some matches not from the start, then some fuzzy matches (these exist because the completion list is computed on `G`).

If the completion list is computed from `GLM`, then it looks like

image

randy3k:
How about the following situation?

```
Abc <- 1
aBc <- 2
```

What do `Ab` and `aB` trigger?

renkun-ken:
image

image

randy3k:
Thank you for working on it. Didn't have the time to test it.

renkun-ken:
> `ab` will trigger both abc and bca?

Currently, no. I only switch from `startWith(x, token)` to `grepl(token, x, ignore.case = TRUE)`. The completions are case-insensitive and should include the token (not necessarily from the start). For example,

```
abc <- 1
cab <- 2
```

`Ab` should return both `abc` and `cab`, but `Ca` should only return `cab`.

randy3k:
I am a bit confused why for example `as.table` is there when `ab` is typed. The a the b in `as.table` are not consecutive. 

Perhaps it was because the list was generated by the server when `a` was typed. And when `b` was typed, VSCode does not query the server to speed up things and only do simple filtering?

renkun-ken:
When `a` is pressed, the completion request is sent, and it got all results that contain `a` or `A`, and `isIncomplete = false`. Then the following typing does not trigger completion request and then vscode does the filtering and sorting according to the token under cursor.

randy3k:
Thank you for the explanation. Totally forgot this setting. It makes total sense.

randy3k (approved):

PR #411, opened 2021-04-06
fix raw string detection

### LINKED ISSUE #410: Completion not working before close parenthesis
Consider the following code

```r
fun <- function(x, y, z) {
  paste0(x, y, z)
}

fun("free"|)
```

Inserting `,` at `|` does not trigger the signature help of `fun(x, y, z)`, and typing e.g. `z` does not get any completion item. However, no error is observed from the logging of languageserver. Also, replacing `"free"` to some other strings, the problem might disappear.

randy3k:
close #410 

- need to cherrypick to release/0.3.10

### DIFF
--- a/src/fsm.c
+++ b/src/fsm.c
@@ -35,8 +35,6 @@ void fsm_feed(fsm_state* state, const char c) {
                 return;
             } else {
                 state->raw_parse_state = 0;
-                state->double_quoted = 0;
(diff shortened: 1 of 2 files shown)


randy3k:
It was caused by the `r` in '`free'`` wrongly triggering the raw string state.

renkun-ken (approved):
LGTM

PR #413, opened 2021-04-08
Faster completion

### LINKED ISSUE #412: Improve completion performance
When I'm dealing with numerous large documents with numerous packages attached, the completion could become laggy as the first keystroke triggers the completion request which more than 2000 items are returned but most are useless.

I think we could truncate the completion list to a number and use `isIncomplete = TRUE` so that next keystroke could still trigger completion request but this time it narrows down the range often significantly. This would probably improve the performance of completion.

renkun-ken:
Closes #412 

This PR improves the performance of completion by limiting the max number of completions to return to the client and utilizing the `isIncomplete` feature.

If the completion list contains more than `nmax` (by default, 200) items, then the list is sorted by whether it starts with `token` and the `sortText` of the item, and the top `nmax` items are returned with `isIncomplete = TRUE` so that subsequent keystrokes will still trigger completion requests until the completion list is narrowed down to no greater than `nmax`.

This largely avoids too much traffic from the response to the completion request triggered by the first keystroke, which could potentially match too many results.

Old behavior (on my machine working in languageserver project):

1. Type `a`
2. 1619 matches returned to client with `isIncomplete=FALSE`, and subsequent keystrokes do not trigger completion request.

In this case, a notable delay could be observed on first keystroke of a token.

New behavior:

1. Type `a`
2. 1619 matches reduced to 200 returned to client with `isIncomplete=TRUE`.
3. Type `s`
4. 288 matches reduced to 200 returned to client with `isIncomplete=TRUE`
5. Type `.`
6. 119 matches returned to client with `isIncomplete=FALSE`, and subsequent keystrokes do not trigger completion request.

In this case, the traffic between server and client is reduced and the delay on first keystroke is much reduced.

### DIFF
--- a/R/completion.R
+++ b/R/completion.R
@@ -520,8 +520,12 @@ completion_reply <- function(id, uri, workspace, document, point, capabilities) 
 
     if (length(completions) >= nmax) {
         isIncomplete <- TRUE
-        sort_text <- vapply(completions, "[[", character(1), "sortText")
-        completions <- completions[order(sort_text)][seq_len(nmax)]
+        label_text <- vapply(completions, "[[", character(1), "label")
+        completions <- completions[startsWith(label_text, token)]
+        if (length(completions) >= nmax) {
+            sort_text <- vapply(completions, "[[", character(1), "sortText")
+            completions <- completions[order(sort_text)][seq_len(nmax)]
+        }
     } else {
         isIncomplete <- FALSE
     }
--- a/R/completion.R
+++ b/R/completion.R
@@ -456,6 +456,8 @@ completion_reply <- function(id, uri, workspace, document, point, capabilities) 
             )))
     }
 
+    nmax <- getOption("languageserver.max_completions", 100)
+
     t0 <- Sys.time()
     snippet_support <- isTRUE(capabilities$completionItem$snippetSupport) &&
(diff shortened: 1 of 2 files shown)


R/completion.R
```diff
@@ -499,10 +501,32 @@ completion_reply <- function(id, uri, workspace, document, point, capabilities)
         )
     }
 
+    init_count <- length(completions)
+    nmax <- getOption("languageserver.max_completions", 200)
```
randy3k, line 505: I think this setting name should be more explicit, it gives users a feeling that this is the absolute maximum number of items.

renkun-ken: It is indeed the max number of completion items to return to the client. The results will be truncated to this number if it is more than that. Am I misunderstanding your meaning?

randy3k: Ya, you are correct. I was confused. Sorry for the noise.

R/completion.R
```diff
@@ -499,10 +501,32 @@ completion_reply <- function(id, uri, workspace, document, point, capabilities)
...
+
+    if (length(completions) > nmax) {
```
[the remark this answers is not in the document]
renkun-ken: Yes, you are right.

randy3k (approved):

PR #416, opened 2021-04-10
Definition provider works with file path

### LINKED ISSUE #415: definitionProvider should work with file path
In [C/C++ extension](https://github.com/Microsoft/vscode-cpptools), "Go to definition" works with compiler directives so that it opens the file specified in the following include directive:

```cpp
#include <iostream>
```

I guess it makes sense to open file in the definitionProvider if the underlying token is `STR_CONST` and refers to a file path. If it is added, then in vscode, we not only could "Go to definition" to a file, we could also "Peek definition" of a file path to view the file content in the middle of the editor tab, or press "option" + mouse hover to preview the file content.

This might require that the file be a text file rather than binary file though.

renkun-ken:
Closes #415 

This PR makes `STR_CONST` token which is either an absolute path or a path relative to workspace `rootPath`.

It also checks if the file is a text file by reading its first 1000 bits  into a raw vector using `readBin` and see if ~~`rawToChar` succeeds~~ it is a text file by detecting its text encoding with `{stringi}` functions.

image

### DIFF
--- a/R/definition.R
+++ b/R/definition.R
@@ -12,7 +12,7 @@ definition_xpath <- paste(
 #' writes the function definition to a temporary file and returns that
 #' as the location.
 #' @keywords internal
-definition_reply <- function(id, uri, workspace, document, point) {
+definition_reply <- function(id, uri, workspace, document, point, rootPath) {
 
     token_result <- document$detect_token(point)
     resolved <- FALSE
@@ -56,6 +56,29 @@ definition_reply <- function(id, uri, workspace, document, point) {
                         resolved <- TRUE
                     }
                 }
+            } else if (token_name == "STR_CONST") {
+                str_line1 <- as.integer(xml_attr(token, "line1"))
+                str_line2 <- as.integer(xml_attr(token, "line2"))
+                if (str_line1 == str_line2) {
+                    str_col1 <- as.integer(xml_attr(token, "col1"))
+                    str_col2 <- as.integer(xml_attr(token, "col2"))
+                    str_expr <- substr(document$content[str_line1], str_col1, str_col2)
+                    str_text <- tryCatch(as.character(parse(text = str_expr, keep.source = FALSE)),
+                        error = function(e) NULL)
+                    if (is.character(str_text)) {
+                        path <- fs::path_abs(str_text, rootPath)
+                        if (file.exists(path) && !dir.exists(path) && is_text_file(path)) {
+                            result <- list(
+                                uri = path_to_uri(path),
+                                range = range(
+                                    start = position(0, 0),
+                                    end = position(0, 0)
+                                )
+                            )
+                        }
+                    }
+                }
+                resolved <- TRUE
             } else {
                 resolved <- TRUE
             }
--- a/R/utils.R
+++ b/R/utils.R
@@ -692,5 +692,15 @@ format_file_size <- function(bytes) {
 
 is_text_file <- function(path, n = 1000) {
     bin <- readBin(path, "raw", n = n)
-    stringi::stri_enc_isutf8(bin)
+    is_utf8 <- stringi::stri_enc_isutf8(bin)
+    if (is_utf8) {
+        return(TRUE)
+    } else {
+        result <- stringi::stri_enc_detect(bin)[[1]]
+        conf <- result$Confidence[1]
+        if (identical(conf, 1)) {
+            return(TRUE)
+        }
+    }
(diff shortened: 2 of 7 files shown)


renkun-ken:
I have no idea why the code coverage fails while the builds succeed. The error looks a bit weird

```
actual vs expected
- "file:///var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T//RtmpEg9DcE/var.R"
+ "file:///var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/RtmpMKNtQM/file92e2f824109.R"
```

where the actual `uri` is not even a random filename.

R/definition.R
```diff
@@ -56,6 +56,29 @@ definition_reply <- function(id, uri, workspace, document, point) {
...
+            } else if (token_name == "STR_CONST") {
+                str_line1 <- as.integer(xml_attr(token, "line1"))
+                str_line2 <- as.integer(xml_attr(token, "line2"))
+                if (str_line1 == str_line2) {
+                    str_col1 <- as.integer(xml_attr(token, "col1"))
+                    str_col2 <- as.integer(xml_attr(token, "col2"))
+                    str_expr <- substr(document$content[str_line1], str_col1, str_col2)
+                    str_text <- tryCatch(as.character(parse(text = str_expr, keep.source = FALSE)),
+                        error = function(e) NULL)
+                    if (is.character(str_text)) {
```
randy3k, line 68: do you really need to use `parse` here? `as.character(expression("abc"))` seems to work.

R/utils.R
```diff
@@ -689,3 +689,9 @@ format_file_size <- function(bytes) {
     obj_size <- structure(bytes, class = "object_size")
     format(obj_size, units = "auto")
 }
+
+is_text_file <- function(path, n = 1000) {
+    bin <- readBin(path, "raw", n = n)
+    result <- tryCatch(rawToChar(bin), error = function(e) NULL)
```
randy3k, line 695: I don't think it will work.
`rawToChar(as.raw(c(1:255)))` basically converts all bytes.

R/definition.R
renkun-ken: It is used to handle raw strings where the text is like `r"{hello}"`.

R/utils.R
renkun-ken: It is in fact a detection of `\0` to cause error based on the assumption that a binary file is quite likely to have `\0` in it.

randy3k: Not necessary true. As we only support UTF-8 files, maybe better to use `stringi::stri_enc_isutf8(bin)`?

renkun-ken: You are right, `stri_enc_isutf8` looks better.

randy3k: Ai, it doesn't work well if you accidentally cut of an unicode at the 1000 byte.
```
stringi::stri_enc_isutf8(charToRaw("中")[1:2])
```

Though it should not be a big concern most of the time.

renkun-ken: What about retry with n-1 at most 3 times of detection. This should work if the 1000 bytes cut in the middle of the a unicode character.

renkun-ken: Looks like a more robust way is to check `stringi::stri_enc_isutf8` first, and then fallback to `stringi::stri_enc_detect` so that if the top encoding has confidence=1 then it is almost always a text file. In this case, I guess we could accept other text encodings if it is so confidently guessed so.

I test with a variety of text and binary files and this approach seems to always deliver the correct result.

randy3k (approved):

randy3k:
@renkun-ken 
Please don't merge it now, I want to investigate why the coverage test failed.

renkun-ken:
@randy3k Would you mind I switch the coverage to using ubuntu. It works as expected: https://github.com/renkun-ken/languageserver/runs/2359742236?check_suite_focus=true.

randy3k:
I chose macOS because travis coverage test is in Linux. However, the macOS build is causing troubles here, so it make senses to switch to Linux for now. WIth that said, it would be interesting to understand why it failed in macOS.

renkun-ken:
> I chose macOS because Travis coverage test is in Linux. However, the macOS build is causing troubles here, so it makes senses to switch to Linux for now.

Sure, I'll merge that switch then.

> With that said, it would be interesting to understand why it failed in macOS.

Exactly, I did a little digging but no conclusion so far.

PR #431, opened 2021-05-27
Use a function call as the completer in on-type-formatting

### LINKED ISSUE #430: Support pipe operator in onTypeFormatting
Currently, we use a trick of appending a zero to complete a partial expression so that it is made stylizable by `styler`.

However, this trick no longer applies to the newly introduced pipe syntax in R 4.1 where it is syntactically incorrect to pipe to anything other than an explicit function call.

Previously if we hit "enter" in the end of the following line

```r
1:10 %>%
```

then onTypeFormating will format the following code

```r
1:10 %>%
0
```

with proper indention, it becomes

```r
1:10 %>%
  0
```

But the same does not work with

```r
1:10 |>
0
```

as it is not accepted by the parser.

Now styler has supported pipe expressions via https://github.com/r-lib/styler/pull/803.

renkun-ken:
Closes #430 

This PR changes the completer from `0` to `f()` so that the pipe syntax could be supported.

### DIFF
--- a/R/formatting.R
+++ b/R/formatting.R
@@ -126,17 +126,17 @@ range_formatting_reply <- function(id, uri, document, range, options) {
 on_type_formatting_reply <- function(id, uri, document, point, ch, options) {
     content <- document$content
     end_line <- point$row + 1
-    use_zero <- FALSE
+    use_completer <- FALSE
     if (ch == "\n") {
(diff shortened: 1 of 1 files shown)


randy3k (approved):
LGTM

PR #437, opened 2021-06-06
check if lintr is new enough

randy3k:
> I just realized that there is no roadmap for the release of new version of lintr. Perhaps we should include your original trick of checking the text parameter for now? Otherwise we won't know how much time they need to make a new release.

related: #284 

I have decided to use `packageVersion` at the end.

### DIFF
--- a/R/diagnostics.R
+++ b/R/diagnostics.R
@@ -5,6 +5,17 @@
 #' @name diagnostics
 NULL
 
+
+#' Check if lintr is 2.0.1.9000 above. A number of features relies on it.
+#' @noRd
+lintr_is_new_enough <- function() {
+    return(
+        utils::packageVersion("lintr") >= "2.0.1.9000" &&
+            "text" %in% methods::formalArgs(lintr::lint)
+    )
+}
+
+
 DiagnosticSeverity <- list(
     Error = 1,
     Warning = 2,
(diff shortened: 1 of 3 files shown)


R/diagnostics.R
```diff
@@ -5,6 +5,14 @@
 #' @name diagnostics
 NULL
 
+
+#' Check if lintr is 2.0.1.9000 above. A number of features relies on it.
+#' @noRd
+lintr_is_new_enough <- function() {
+    return(compare_version(as.character(utils::packageVersion("lintr")), "2.0.1.9000") >= 0)
```
renkun-ken, line 12: `utils::packageVersion()` returns a `package_version` object which implements version comparison out of the box. Is there a reason not to directly use `>=` but implements another `compare_version` here?

randy3k: Oh, didnt know that. I thought it is only a string. It seems that I have just reinvented the wheel again. Will fix it soon.

renkun-ken:
Also, the lintr version `2.0.1.9000` could be either before or after https://github.com/jimhester/lintr/pull/503 was merged. Therefore, I was using https://github.com/REditorSupport/languageserver/pull/284/files/4d42f4a63d79784340172a484cd6447db6ecbe97#diff-5acd9f0d2237a48b4e62aa0d9ee52a517602b6d4d118ed56187f668f94e19011R93 to detect if `text=` argument is available rather than using a version number.

randy3k:
I have made some changes. Please review again.

R/diagnostics.R
```diff
@@ -5,6 +5,17 @@
 #' @name diagnostics
 NULL
 
+
+#' Check if lintr is 2.0.1.9000 above. A number of features relies on it.
+#' @noRd
+lintr_is_new_enough <- function() {
+    return(
+        utils::packageVersion("lintr") >= "2.0.1.9000" &&
+            "text" %in% methods::formalArgs(lintr::lint)
```
renkun-ken, line 14: The check failed as we use `methods::` which is not declared in `Import`. We could simpliy use `names(formals(lintr::lint))` without having to import methods as before?

renkun-ken (approved):
LGTM

PR #438, opened 2021-06-06
[WIP] an option to disable rich documentation

randy3k:
it is needed for editor with less fancy markdown renderer.

Shall we actually turn it off as default and only provide the rich markdown if client to specify via
[MarkdownClientCapabilities](
https://microsoft.github.io/language-server-protocol/specifications//specification-3-17/#markdownClientCapabilities)? I imagine vscode-r-lsp could enable it by default.

### DIFF
--- a/R/workspace.R
+++ b/R/workspace.R
@@ -154,8 +154,9 @@ Workspace <- R6::R6Class("Workspace",
                 } else {
                     result <- NULL
 
-                    if (requireNamespace("rmarkdown", quietly = TRUE) &&
-                        rmarkdown::pandoc_available()) {
+                    if (isTRUE(getOption("languageserver.rich_documentation", TRUE)) &&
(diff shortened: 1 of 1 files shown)


randy3k:
Actually, now I am thinking that we should less depend on settings via `.Rprofile` but utilizing the `workspace/didChangeConfiguration`. So ignore this PR for now, I will come back with a better solution.

randy3k:
We have the following settings

```
languageserver.snippet_support
languageserver.max_completions
languageserver.lint_cache
languageserver.formatting_style
languageserver.server_capabilities
languageserver.link_file_size_limit
```

I think we should find a way to consistently expose them to `workspace/didChangeConfiguration` as well as via `options` from `.Rprofile`.

randy3k:
closed in flavor of #439

PR #447, opened 2021-07-01
Handle `workspace/didChangeWatchedFiles`

### LINKED ISSUE #446: Handle `workspace/didChangeWatchedFiles`
There are some scenarios where some R files are created, changed, or removed outside editor operations which is not sent from client to server via lsp requests.

* Git checkout
* Rcpp/cpp11 generate R wrapper file for exported/registered C++ functions

In these cases, language server is not aware of the changes of the files, and thus will not update the workspace so that symbol definition and etc will not update accordingly until user explicitly open the changed files.

Therefore, we should handle [`workspace/didChangeWatchedFiles`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_didChangeWatchedFiles).

renkun-ken:
Closes #446 

This PR implements `workspace/didChangeWatchedFiles` so that in a package workspace whenever files in `R` folders are changed, the created or modified files will be reloaded, and the removed files are dropped. As a result, the workspace is always up-to-date.

Note that the changes in open documents will trigger `workspace/didChangeWatchedFiles` too, but the client will send requests to properly handle them in the first place. Therefore, we explicitly ignore these open documents here.

- [x] Implement `workspace/didChangeWatchedFiles`
- [x] Test cases


renkun-ken:
@randy3k Any comments?

randy3k:
LGTM. I think at some point we will need to look into supporting non-package projects, eg a shiny or plumber app.

randy3k (approved):

R/handlers-textsync.R
```diff
@@ -103,6 +106,9 @@ text_document_did_close <- function(self, params) {
         diagnostics_callback(self, uri, NULL, list())
         self$workspace$documents$remove(uri)
         self$workspace$update_loaded_packages()
+    } else if (self$workspace$documents$has(uri)) {
+        doc <- self$workspace$documents$get(uri)
+        doc$did_close()
```
randy3k, line 111: To simplify the logic, could we move these line to line 103?

randy3k: You merged quick :) Never mind.

renkun-ken: Done via 354b8d8

PR #458, opened 2021-07-19
Require styler >= 1.5.1

### LINKED ISSUE #457: Minimal {styler} version requirement
{styler} won't ever bring up interactive prompt to create a cache (as requested by {R.cache}) as of version 1.5.1. For this reason, the corresponding README explanation becomes obsolete if you add a minimal version requirement on {styler} (`1.5.1`), which I suggest to do.
Also see https://github.com/r-lib/styler/pull/819.

renkun-ken:
Closes #457 

This PR makes the package require styler >= 1.5.1 so that the default cache root path (`~/.cache/R/R.cache`) is global to the user rather than per session. No prompt required to use a global cache root. Therefore, there is no need to mention the cache issue in the README.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,7 +32,7 @@ Imports:
     repr (>= 1.1.0),
     roxygen2 (>= 7.0.0),
     stringi (>= 1.1.7),
-    styler (>= 1.4.1),
+    styler (>= 1.5.1),
     tools,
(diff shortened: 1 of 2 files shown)


randy3k:
Do you think we are ready for another release?

renkun-ken:
Sure, please go ahead!

PR #464, opened 2021-08-06
Fix several bugs

### LINKED ISSUE #463: Bug when hovering variables
**Describe the bug**
When hover on a variable defined before, it shows the definition of the variable. But in some cases, definition isn't correct. For example, paste the following code in the vscode.
```
for (ll in 1:3){
 p = ll
 I = array(0:0, dim=c(p,p))
}
```
Then hover on the variable `p`. The correct behavior should be popping up a panel showing `p=ll`, but it shows `for (ll in 1:3){`. A screenshot is below.

image

renkun-ken:
Closes #463 

Hovering on the left part of a symbol might capture the symbol token with `document$detect_token()` while the XPath captures the token on the left of the cursor, e.g. an operator, and thus shows the workspace definition of the captured token rather than the scope definition.

This PR:

* Ignores the operator captured by XPath.
* Replaces all `\t` with a space in `parse_document()` to make the width of tab consistent between LSP and `getParseData()`.
* Improves the XPaths of hover, definition, signature, and scope completion to support code with semi-colons where nested `exprlist` nodes appear in the XML syntax tree.

### DIFF
--- a/R/document.R
+++ b/R/document.R
@@ -402,6 +402,8 @@ parse_document <- function(uri, content) {
     if (is_rmarkdown(uri)) {
         content <- purl(content)
     }
+    # replace tab with a space since the width of a tab is 1 in LSP but 8 in getParseData().
+    content <- gsub("\t", " ", content, fixed = TRUE)
     expr <- tryCatch(parse(text = content, keep.source = TRUE), error = function(e) NULL)
     if (!is.null(expr)) {
         parse_env <- function() {
--- a/R/completion.R
+++ b/R/completion.R
@@ -313,17 +313,17 @@ workspace_completion <- function(workspace, token,
 }
 
 scope_completion_symbols_xpath <- paste(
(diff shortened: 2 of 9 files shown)


randy3k:
Maybe adding a test for it?

R/document.R
```diff
@@ -402,6 +402,8 @@ parse_document <- function(uri, content) {
     if (is_rmarkdown(uri)) {
         content <- purl(content)
     }
+    # replace tab with a space since the width of a tab is 1 in LSP but 8 in getParseData().
```
randy3k, line 405: cf: https://github.com/wch/r-source/blob/8a0aaaa4feac8a97c211771bda8fd1454144ee3c/src/main/gram.y#L597

renkun-ken: Thanks for digging into the source!

renkun-ken:
Test cases are added. Any other suggestions?

randy3k:
Thanks for working on it. LGTM.

randy3k (approved):

PR #467, opened 2021-08-11
Fix hover on function defined with `EQ_ASSIGN`

### LINKED ISSUE #466: Documentation for functions assigned with `=` differs from functions assigned with `<-`
It seems that definitions for functions are not extracted properly when those are assigned with eq. Is this intended? This has an annoying effect for example when hovering over functions with signatures spread to multiple lines:

hover-func-r-eq

renkun-ken:
Closes #466 

This PR fixes the XPath to extract function signature from the definition with `EQ_ASSIGN` by matching `expr/LEFT_ASSIGN` and `expr_or_assign_or_help/EQ_ASSIGN` in the XML parse tree.

### DIFF
--- a/R/hover.R
+++ b/R/hover.R
@@ -50,7 +50,7 @@ hover_reply <- function(id, uri, workspace, document, point) {
                     if (length(all_defs)) {
                         last_def <- all_defs[[length(all_defs)]]
                         def_func <- xml_find_first(last_def,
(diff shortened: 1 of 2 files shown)


randy3k:
LGTM. You have became an expert of xpath 😄.

randy3k (approved):

PR #489, opened 2021-10-18
bump to v0.3.12

randy3k:


renkun-ken:
Should we put imported packages in the namespace as noted at <https://cran.r-project.org/web/checks/check_results_languageserver.html>?

PR #498, opened 2021-11-21
Unpublish diagnostics when disabled

### LINKED ISSUE #497: Disabling linting has no effect
In VS Code, disabling `r.lsp.diagnostics` seems to have no effect and the following error appears in the log:

```
[2021-11-21 13:07:48.202] internal error: Error: argument "params" is missing, with no default
Call: dispatch(self, params)
Stack trace:
1: dispatch(self, params)

[2021-11-21 13:07:48.203] received:  Content-Length: 185

[2021-11-21 13:07:48.216] received payload.

[2021-11-21 13:07:48.216] handling notification:  workspace/didChangeConfiguration

[2021-11-21 13:07:48.216] settings  {
  "enabled": true,
  "args": [],
  "debug": true,
  "diagnostics": false,
  "lang": "",
  "use_stdio": false
}

```

renkun-ken:
Closes #497 

This PR unpublishes diagnostics of all workspace documents when diagnostics is disabled from client settings.

Also, this PR corrects the signature of `protocol_set_trace` to `protocol_set_trace(self, params)`. This handler turns on/off `debug` previously, which might cause conflicts with `r.lsp.debug` since if it is `true` but the language server trace is not turnoed on, then on the server side the debug will be enabled on startup but then disabled due to `setTrace` notification says `off`.

I add a new `trace` setting parallel with `debug` as well as `logger$trace` so that they have no conflicts.

`logger$info` works if either `debug` or `trace` is on, and `logger$trace` only works if `trace` is on.

### DIFF
--- a/R/handlers-general.R
+++ b/R/handlers-general.R
@@ -6,7 +6,7 @@
 on_initialize <- function(self, id, params) {
     trace <- params$trace
     if (!is.null(trace) && trace %in% c("messages", "verbose")) {
-        lsp_settings$set("debug", TRUE)
+        lsp_settings$set("trace", TRUE)
     }
 
     logger$info("session: ", list(
--- a/R/handlers-protocol.R
+++ b/R/handlers-protocol.R
@@ -1,8 +1,8 @@
-protocol_set_trace <- function(self, id, params) {
+protocol_set_trace <- function(self, params) {
     value <- params$value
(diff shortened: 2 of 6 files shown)


randy3k:
Thank you. LGTM.

randy3k (approved):

PR #509, opened 2021-12-28
Document formatting ensures trailing empty line

### LINKED ISSUE #462: "No newline at end of file" after formatting file in VS Code
1. Create a file `bug.R` containting this line: `styler::style_file("bug.R")` (no newline at end of file).
2. Run it using `Rscript bug.R`
3. Note that a newline at the end of file has been introduced. So this seems to be tidyverse style.
4. Note also that the GitHub web-interface file editor introduces a newline at the end of the file - there is no way to even avoid it.
5. `.Rprofile` ignores the last line if it does not end in a newline, see https://yihui.org/en/2018/04/rprofile-trailing-newline/

I think the above establishes that a file should end in a newline character.

Now, the R language server seems to remove that newline character - inconsistently!

hTLPI9Kei2

Note especially how with `"editor.formatOnSave": true`, it takes me 5 tries to reproduce the problem.

renkun-ken:
Closes #462 

In document formatting, a trailing new line is ensured in the end of the new text.

### DIFF
--- a/R/formatting.R
+++ b/R/formatting.R
@@ -14,7 +14,7 @@ get_style <- function(options) {
 #' specified style.
 #'
 #' @noRd
-style_text <- function(text, style, indention = 0L) {
+style_text <- function(text, style, indention = 0L, trailing_empty_line = FALSE) {
     new_text <- tryCatch(
         styler::style_text(
             text,
@@ -26,6 +26,11 @@ style_text <- function(text, style, indention = 0L) {
     if (inherits(new_text, "error")) {
         logger$info("formatting error:", new_text$message)
         return(NULL)
+    }
+    if (isTRUE(trailing_empty_line)) {
+        if (new_text[[length(new_text)]] != "") {
+            new_text <- c(new_text, "")
+        }
     }
     paste0(new_text, collapse = "\n")
 }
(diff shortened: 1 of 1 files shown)


lorenzwalthert:
I read this as *if the last line is not empty, add one* (if `trailing_empty_line`). However, if you have already 5 empty lines, that won't remove 4. You can also just c/p the `ensure_last_n_empty()` from {styler}:
```r
ensure_last_n_empty <- function(x, n = 1) {
  if (all(x == "")) {
    return("")
  }
  x <- c(x, "", "")
  x <- x[seq(1, length(x) - which(rev(x) != "")[1] + 1L)]
  c(x, rep("", n))
}
```

``` r
styler:::ensure_last_n_empty("hi")
#> [1] "hi" ""
styler:::ensure_last_n_empty(c("hi", ""))
#> [1] "hi" ""
styler:::ensure_last_n_empty(c("hi", "", ""))
#> [1] "hi" ""
```

<sup>Created on 2021-12-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

renkun-ken:
Thanks for pointing to `styler:::ensure_last_n_empty()`. I actually tried it, but I wonder if we should regard removing trailing new lines as part of styler rather than languageserver?

The current release does not remove trailing empty lines, and the latest dev removes all trailing empty lines. I think it might make some sense to leave it to styler to determine whether it is appropriate to remove all trailing empty lines? On languageserver side, we only need to ensure there is at least a trailing empty line to make things work (e.g. `.Rprofile`).

lorenzwalthert:
Ok, makes sense. {styler} does remove these additional lines, so with the next CRAN release, people will have this functionality anyways. Thanks for looking into it @renkun-ken, appreciate it that you help fix a problem that is fundamentally not a {languageserver}, but a {styler} problem. 👍

randy3k (approved):

PR #532, opened 2022-03-23
Added an entry for BBEdit

siegel:
…which has LSP support built in beginning with version 14.0 (and support for the R language server is preconfigured).


renkun-ken (approved):
Thanks!

PR #534, opened 2022-03-28
convert the table of arguments to the sections of arguments with Lua filter

### LINKED ISSUE #533: Some of the tables of arguments are too complex to be represented by GFM
As @eitsupi has raised issue in https://github.com/REditorSupport/vscode-R/issues/927 , there are some cases where `Arguments` become VS Code ignores the section if the section remains to be an HTML table due to its complex content. See details in the link.

To avoid this, we need to convert the table to some alternative format.
I would suggest the following style.

````
### Arguments

#### `Arg1`:

Description of  Arg1

#### `Arg2`:

Description of Arg2.

This may contain multiple paragraphs.

- and some list

or

```
code blocks
```
````

atusy:
closes #533

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: languageserver
 Title: Language Server Protocol
-Version: 0.3.12
+Version: 0.3.12.9000
 Date: 2021-10-18
 Authors@R: 
     c(person(given = "Randy",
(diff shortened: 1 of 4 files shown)


atusy:
With this PR, complex arguments becomes sections rather than tables.

image

It even allows bullet lists.

image

If arguments are simple enough, tables remain.

renkun-ken (approved):
Works nicely! Thanks!

atusy:
Thank you, too!!

PR #541, opened 2022-04-18
do not pop the unbalance brackets in the current row

### LINKED ISSUE #530: Signature help not show up in if-else branch
Consider the following code:

```r
if (sum()) {

} else if (sum()) {

}
```

In the `if (sum())` branch, the signature help of `sum()` works while it does not work in the `else if (sum())` branch when I type `sum(`.

randy3k:
close #530


renkun-ken (approved):
Works nicely. LGTM.

Thanks!

PR #552, opened 2022-05-23
minor refactoring: simplify the version number handling

eitsupi:
I believe this simple expression can be used because the `numeric_version` class object can compare version numbers separated by periods.


randy3k:
Thanks, didn't know it.

PR #565, opened 2022-07-25
fix NA introduced into position

### LINKED ISSUE #564: NA introduced into position when opening curly braces
The language server introduces NA when code like the following occurs

```R
function() {
  test <- print(
}
```

This causes for the Neovim lsp interface to error out. In this like shiny this is a very common thing to find. 

My fix was pretty easy, I went to the `code_point_to_unit` function and added the following:

```R
result[is.na(pts)] <- m
```

This fixes the issue and does not seem to break any other behaviour. This is a pretty minor fix. Hope it can get merged!

andyquinterom:
Resolves #564

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -424,7 +424,7 @@ code_point_to_unit <- function(line, pts) {
     n <- length(offsets)
     m <- offsets[n]
     result[pts >= n] <- m
-    result[is.infinite(pts)] <- m
+    result[!is.finite(pts)] <- m
     result
 }
 


R/utils.R
```diff
@@ -425,6 +425,7 @@ code_point_to_unit <- function(line, pts) {
     m <- offsets[n]
     result[pts >= n] <- m
     result[is.infinite(pts)] <- m
+    result[is.na(pts)] <- m
```
renkun-ken, line 428: Looks like we could just replace 

```r
result[is.infinite(pts)] <- m
```

with

```r
result[!is.finite(pts)] <- m
```

and both `Inf` and `NA` cases are handled.

randy3k: agree

andyquinterom: Sounds good to me

andyquinterom:
Fixed!

renkun-ken (approved):
LGTM

Issue #567, opened 2022-08-12
lsp sees things that lintr doesnt


jam1015:
I might have a function defined in  a separate file of functions that I source in the script.  the linter warns about `no global function definition`, but if I use the jump to definition capability it takes me exactly where I need to go. so the lsp knows the function exists but the linter provided by `lintr` doesn't.  I might turn off that linter capability in my `.lintr` file but I wish it worked properly for cases where the function is actually not available.

renkun-ken:
Are you using the latest lintr (3.0.x)?

psanker:
I also have this issue and commented here: https://github.com/REditorSupport/vscode-R/issues/729#issuecomment-1218416818

I noticed that on https://github.com/REditorSupport/languageserver/blob/9f0e29af3d6c8e4e61eb672d9b04d4ec2ca06706/R/diagnostics.R#L87
languageserver only calls `lintr::lint()`, but according to the lintr 3.0.x docs "‘lint()’ lints a single file." I know the LSP previously had a distinction between `lintr::lint()` and `lintr::lint_package()`; is that no longer needed?

jam1015:
@renkun-ken  I am using the latest version of lintr.  I see the links posted by @psanker but I'm not an advanced-enough user to understand them.  for context, using the `languageserver` package in neovim with coc.

renkun-ken:
@jam1015 You might want to try `remotes::install_github("REditorSupport/languageserver#568")` and see if it works for you.

katrinabrock:
It doesn't seem like lsp is loading my package before checking for object definitions. Is there something I need to do to turn this on?

I have `lintr_3.2.0`, `pkgload_1.4.0`, `languageserver_0.3.16` installed. REditorSupport extension version: `0.1.3`

PR #568, opened 2022-08-18
Provide package symbols before lint in diagnostics

renkun-ken:
Closes https://github.com/REditorSupport/vscode-R/issues/729.

Also addresses https://github.com/REditorSupport/languageserver/issues/567#issuecomment-1218437202.

`pkgload::load_all()` is called before `lintr::lint()` if the workspace root contains a package.


renkun-ken:
There are known limitations:

* If `pkgload` is not available, it will still lint the single file.
* If there are parse errors in any R file in R folder, or the R files cannot be evaluated properly, `pkgload::load_all()` will fail, and it will lint the single file.

randy3k:
I am curious if it would be too slow for large package. (Though lintr is slow anyway).

renkun-ken:
I tried `system.time(pkgload(compile = FALSE, quiet = TRUE))` with lintr and the timing looks like

```
> system.time(pkgload::load_all(compile = FALSE, quiet = TRUE))                                                   
   user  system elapsed 
  0.772   0.060   0.893 

> system.time(pkgload::load_all(compile = FALSE, quiet = TRUE))                                                   
   user  system elapsed 
  0.358   0.016   0.375 

> system.time(pkgload::load_all(compile = FALSE, quiet = TRUE))                                                   
   user  system elapsed 
  0.258   0.015   0.272 

> system.time(pkgload::load_all(compile = FALSE, quiet = TRUE))                                                   
   user  system elapsed 
  0.254   0.015   0.270 

> system.time(pkgload::load_all(compile = FALSE, quiet = TRUE))                                                   
   user  system elapsed 
  0.262   0.015   0.277 

> system.time(pkgload::load_all(compile = FALSE, quiet = TRUE))                                                   
   user  system elapsed 
  0.233   0.017   0.250 
```

Since the diagnostics runs in reused sessions, the timing might look better after the first time.

randy3k:
> Since the diagnostics runs in reused sessions, the timing might look better after the first time.

Good point.

dgkf:
I'm seeing this issue as well.

> I am curious if it would be too slow for large package. (Though lintr is slow anyway).

I do have some packages where `load_all()` might be slow. I'm not sure how it would compare, but one alternative would be to filter these diagnostic messages from the `lintr` results when a definition is found within the package for the symbol.

renkun-ken:
@dgkf 

>I do have some packages where load_all() might be slow.

Does it always take as much time as the first time for each `load_all()`? Since we reuse existing sessions to perform diagnostics, repeated `load_all()` should take much less time than the first time.

dgkf:
@renkun-ken - it does get faster with each call, you're right. 

I think either way you go with this would be fine, I only mentioned it because it was my first thought when thinking about how to approach this and wanted to make the suggestion in case it simplified things at all. If this feels like the right approach, I totally trust your judgement here.

renkun-ken:
@randy3k Do you think we should opt-in or opt-out such behavior? AFAIK, only `lintr::object_usage_linter` needs such behavior. If user does not use this linter, there's no point to pay a cost for it.

renkun-ken:
Another approach is that we somehow mimic the behavior of `pkgload::load_all()` but only does the minimal required work to make `lintr::object_usage_linter`, e.g. only populate a namespace where package objects are only named objects with no useful/actual values.

randy3k:
There is really no much additional time added. I am happy either way.

renkun-ken:
A notable impact of using `pkgload::load_all()` is that it seems to render high CPU and memory footprints as I type in the editor, an almost each key press triggers a `load_all()` call, which gradually becomes timely but looks quite expensive in terms of the usage of machine resources.

randy3k:
That could be a potential issue of having high CPU and memory usage. Perhaps we should disable it in default and only allow interested users to opt in and test the feature.

renkun-ken:
I just switch to a simpler approach by attaching a shim environment of `globals` defined in the package, i.e. functions and symbols. Since the overhead of `load_all()` looks quite significant, and much of it does not seem necessary.

@dgkf Would you like to try it out and see if it works for you?

dgkf:
> I just switch to a simpler approach by attaching a shim environment of `globals` defined in the package, i.e. functions and symbols, which has almost zero overhead. The overhead of `load_all()` looks quite significant, and much of it does not seem necessary.
> 
> @dgkf Would you like to try it out and see if it works for you?

Sure - I'll give the brancha go tonight.

dgkf:
I tried out the branch and performance seemed pretty good. Nice work!

I did notice that it will still report the "no visible binding" warnings until the file is saved for the first time, even if the names are found in the local source code.

Otherwise it looks great, and having to do a quick save once they pop up is a small inconvenience given how much of an improvement this already is.

renkun-ken:
If the document is open before the package R files are parsed, its diagnostics will report `no visible binding`. If the package is finished parsing, then the warning will disappear on any edit of the file.

In fact, if you edit other files and don't save them, it should still work because language server workspaces work without assuming all files are saved. By constrast, `load_all()` will only load the saved version of the files, instead of modified by unsaved ones.

renkun-ken:
@randy3k I think this PR is ready to merge. Let me know if you have any other suggestions.

R/diagnostics.R
```diff
@@ -113,12 +119,32 @@ diagnostics_callback <- function(self, uri, version, diagnostics) {
 diagnostics_task <- function(self, uri, document, delay = 0) {
     version <- document$version
     content <- document$content
+
+    is_package <- is_package(self$rootPath)
+    globals <- NULL
+
+    if (is_package) {
+        globals <- new.env(parent = emptyenv())
```
randy3k, line 127: Perhaps we should extract these code to make it a proper method in the workspace class? It may be useful in other cases.

randy3k (approved):
This is a smart solution. Tha k you for working on it. Just a small nitpick.

R/diagnostics.R
renkun-ken: Here, `globals` is an environment only to be attached before `lintr::lint()` so that existing (but fake) package symbols could be found and functions are kept as functions.

The structure is so special that it might only fit the purpose of lintr at the moment. I guess we could revisit here if a more general logic is needed somewhere else.

R/document.R
```diff
@@ -362,7 +362,9 @@ parse_expr <- function(content, expr, env, srcref = attr(expr, "srcref")) {
 
                     if (type == "function") {
                         env$functs <- c(env$functs, symbol)
-                        env$formals[[symbol]] <- value[[2L]]
+                        fun <- function() NULL
+                        formals(fun) <- value[[2L]]
```
renkun-ken, line 366: `parse_data` now preserves the functions defined in the documents with only formals, which could make it easier to work with. If a future version of lintr checks if function arguments supplied match the function formals, it should be still working as the formals are preserved.

Also, it make it easier to works with `match.call(fun, expr)` in the future.

R/diagnostics.R
randy3k: Sounds good.

PR #577, opened 2022-10-13
Drop diagnostics on callback if disabled

renkun-ken:
Closes https://github.com/REditorSupport/vscode-R/issues/1196

If the client opens a document before the language server receives settings from the client, diagnostics will be triggered (which is default). If diagnostics is disabled, `diagnostics_callback` will still publish the diagnostics results, which is undesirable.

This PR corrects this behavior by checking diagnostics settings again before publishing results. At this point, it is most likely for the server to have already received the settings from the client.


randy3k (commented):
LGTM. Thanks.

I think a new release is due and we should release an update.

renkun-ken:
Yes, agree.

PR #578, opened 2022-10-13
bump to v0.3.14

randy3k:
Will merge after CRAN accepts the submission.

### DIFF
--- a/R/diagnostics.R
+++ b/R/diagnostics.R
@@ -85,6 +85,7 @@ diagnose_file <- function(uri, content, is_rmarkdown = FALSE, globals = NULL, ca
     }
 
     if (length(globals)) {
+        parent.env(globals) <- environment()
         lints <- with(globals, lintr::lint(path, cache = cache, text = content))
     } else {
         lints <- lintr::lint(path, cache = cache, text = content)
--- a/R/diagnostics.R
+++ b/R/diagnostics.R
@@ -85,12 +85,11 @@ diagnose_file <- function(uri, content, is_rmarkdown = FALSE, globals = NULL, ca
     }
 
     if (length(globals)) {
-        env_name <- "languageserver:globals"
-        attach(globals, name = env_name, warn.conflicts = FALSE)
-        on.exit(detach(env_name, character.only = TRUE))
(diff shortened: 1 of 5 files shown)


randy3k:
@renkun-ken CRAN check doesn't like the use of attach, PTAL.

R/diagnostics.R
```diff
@@ -85,12 +85,12 @@ diagnose_file <- function(uri, content, is_rmarkdown = FALSE, globals = NULL, ca
     }
 
     if (length(globals)) {
-        env_name <- "languageserver:globals"
-        attach(globals, name = env_name, warn.conflicts = FALSE)
-        on.exit(detach(env_name, character.only = TRUE))
+        parent.env(globals) <- environment()
+        lints <- with(globals, lintr::lint(path, cache = cache, text = content))
```
renkun-ken, line 89: What if `globals` contains symbols such as `path`, and cache, and `content`?

renkun-ken: I tried putting `path <- NULL` in a package R file, and the following occurs:

```
Failed to run diagnostics: ! error in callr subprocess
Caused by error in `if (needs_tempfile) { ...`:
! missing value where TRUE/FALSE needed
```

randy3k: Good catch. Let me fix it.

renkun-ken: Looks like the variables must be accessible from the global environment. Looks like `globals()` must be attached and we'll have to walk-around the CRAN policy?

renkun-ken: Just like what callr has done at 

https://github.com/r-lib/callr/blob/5adbb27027c377b3ad0d4b3b471f81b36b02da67/R/hook.R#L7

randy3k: +1

renkun-ken: Already fixed

renkun-ken (approved):
LGTM

R/diagnostics.R
randy3k: Thanks, submitting to CRAN.

randy3k:
It's on the way to CRAN.

PR #589, opened 2022-12-27
Disable linters actions

renkun-ken:
Code actions are improved to support disable relevant linters when diagnostics are present.

image

When a code action of disabling a linter is chosen, its name will be added after `# nolint:` directive if exists.


randy3k (commented):
LGTM, thanks for working on it.

randy3k (approved):

PR #591, opened 2023-01-05
Include all exports in namespace objects

### LINKED ISSUE #590: package::function() autocompletion doesn't include re-exported functions
In RStudio, I can use autocomplete to complete function names such as:

`purrr::set_names()` and `rvest::read_html()`.

These are functions that belong originally to other packages (`rlang` and `xml2` respectively in these examples) but are also exported from `purrr` and `rvest`.

In R in VSCode, if I try to autocomplete `rvest::` I don't get `read_html()` as a dropdown option. I can only get it by starting with `xml2::`. Is there a way that languageserver could read in these borrowed functions as options for autocomplete?

renkun-ken:
Closes #590 

`objects(ns)` does not include all exports such as `rvest::read_html()` (from `xml2`) and `purrr:set_names()` (from `rlang`). This PR makes the `Namespace` object include all namespace objects and all exports (not necessarily live in the same namespace).

### DIFF
--- a/R/namespace.R
+++ b/R/namespace.R
@@ -21,7 +21,7 @@ PackageNamespace <- R6::R6Class("PackageNamespace",
             objects <- union(names(ns), exports)
             private$objects <- sanitize_names(objects)
             is_function <- vapply(private$objects, function(x) {
-                is.function(get(x, envir = ns))
+                is.function(ns[[x]])
             }, logical(1L), USE.NAMES = FALSE)
             is_exported <- private$objects %in% exports
             private$functs <- private$objects[is_function]
--- a/R/namespace.R
+++ b/R/namespace.R
@@ -21,7 +21,7 @@ PackageNamespace <- R6::R6Class("PackageNamespace",
             objects <- union(names(ns), exports)
             private$objects <- sanitize_names(objects)
             is_function <- vapply(private$objects, function(x) {
-                is.function(ns[[x]])
+                is.function(get0(x, ns))
             }, logical(1L), USE.NAMES = FALSE)
             is_exported <- private$objects %in% exports
             private$functs <- private$objects[is_function]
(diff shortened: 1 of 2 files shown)


randy3k:
Oh, you are 4 minutes faster than mine, I have just pushed https://github.com/REditorSupport/languageserver/tree/reexport

R/namespace.R
```diff
@@ -17,12 +17,13 @@ PackageNamespace <- R6::R6Class("PackageNamespace",
         initialize = function(pkgname) {
             self$package_name <- pkgname
             ns <- asNamespace(pkgname)
-            private$objects <- sanitize_names(objects(ns, all.names = TRUE))
+            exports <- getNamespaceExports(ns)
+            objects <- union(names(ns), exports)
+            private$objects <- sanitize_names(objects)
             is_function <- vapply(private$objects, function(x) {
-                obj <- tryCatch(get(x, envir = ns), error = function(e) NULL)
```
randy3k, line 22: I forgot why I added tryCatch here, maybe it is before I added `sanitize_names` to make sure the "objects" are real objects not something like `names<-`.

renkun-ken: Maybe `ns[[x]]` is simpler and will not throw error?

randy3k: Ai, actually it doesn;t work
```
r$> asNamespace("purrr")[['set_names']]
NULL

r$> get('set_names', asNamespace("purrr"))
function (x, nm = x, ...)
{
    mold <- x
    .Call(ffi_set_names, x, mold, nm, environment())
}
<bytecode: 0x7fe85d9f0ea8>
<environment: namespace:rlang>
```

renkun-ken:
> Oh, you are 4 minutes faster than mine, I have just pushed https://github.com/REditorSupport/languageserver/tree/reexport And our commits are basically the same :)

Didn't notice your commits :) Just added a test case for this.

R/namespace.R
renkun-ken: Good catch! I switch to `get0` instead.

PR #608, opened 2023-03-24
Fix regex pattern to match chunk label after comma

heavywatal:
Now '```{r, chunklabel}' is recognized.


renkun-ken (approved):
LGTM Thanks!

renkun-ken:
Some tests failed. Looks like the change in regex breaks some cases.

PR #614, opened 2023-04-20
make sure column is defined

### LINKED ISSUE #609: Error in Neovim: “attempt to compare string with number”
Based on the error stack traceback, this is related to #481. Unfortunately #565 did not fix it, and the error persists in the current HEAD revision (adc2189eff742fec42f3ece6df5c3c33062607b3).

To reproduce the issue, create a file containing the following and open it in Neovim with LSP configured to use ‘languageserver’:

```r
a =
b = 2
```

Here’s the full error message:

```
Error executing vim.schedule lua callback: /opt/local/share/nvim/runtime/lua/vim/diagnostic.lua:431: attempt to compare string with number
stack traceback:
        /opt/local/share/nvim/runtime/lua/vim/diagnostic.lua:431: in function 'add'
        /opt/local/share/nvim/runtime/lua/vim/diagnostic.lua:468: in function 'get_diagnostics'
        /opt/local/share/nvim/runtime/lua/vim/diagnostic.lua:1162: in function 'show'
        /opt/local/share/nvim/runtime/lua/vim/diagnostic.lua:724: in function 'set'
        /opt/local/share/nvim/runtime/lua/vim/lsp/diagnostic.lua:206: in function 'handler'
        /opt/local/share/nvim/runtime/lua/vim/lsp.lua:1049: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
```

randy3k:
an attempt to fix #609

### DIFF
--- a/R/diagnostics.R
+++ b/R/diagnostics.R
@@ -17,6 +17,9 @@ DiagnosticSeverity <- list(
 diagnostic_range <- function(result, content) {
     line <- result$line_number - 1
     column <- result$column_number - 1
+    if (is.null(column) || is.na(column)) {
+        column <- 0
+    }
     text <- if (line + 1 <= length(content)) content[line + 1] else ""
     if (is.null(result$ranges)) {
         cols <- code_point_to_unit(text, c(column, column + 1))
--- a/R/utils.R
+++ b/R/utils.R
@@ -276,8 +276,8 @@ code_point_from_unit <- function(line, units) {
 #'
 #' @noRd
 code_point_to_unit <- function(line, pts) {
-    if (!nzchar(line)) return(pts)
     pts[pts < 0] <- 0
+    if (!nzchar(line)) return(pts)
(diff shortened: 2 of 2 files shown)


renkun-ken (approved):
LGTM

randy3k:
Let wait and see if @klmr could confirm that it fixes the issue.

klmr:
Unfortunately this does *not* fix the issue. And when logging the value of `column` I can see that it is `-1` with the minimal test case from #609.

Of course the code could simply be modified to test for values ≤ 0 but maybe we should figure out why the value is negative in the first place. This seems to come directly from `lintr::lint`, correct?

randy3k:
In this case, the bug is actually triggered by `code_point_to_unit` which assumes that offset is always >=0. Hopefully it will fix the issue.

klmr:
Hallelujah, I think that fixes it. And it even returns the correct (rather than just an arbitrary) position for the diagnostic to the language client, I believe.

randy3k:
Great, let's merge it.

PR #615, opened 2023-04-23
Skip incomplete lines in on-type-formatting

renkun-ken:
This PR skips incomplete lines in the backward parsing in on-type-formatting.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1,3 +1,17 @@
+# ?Syntax
+# : can indicate :: :::
+# - can indicate <-
+# > can indicate |>
+binary_opts <- c(
+    ":", "\\$", "@", "\\^",
+    "%[^#]*%", "\\+", "-", "\\*", "/",
+    "<", ">", "=", "!", "&", "\\|", "~",
+    "\\?"
+)
+binary_opts_regex <- paste0(binary_opts, collapse = "|")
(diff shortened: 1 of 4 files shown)


R/utils.R
```diff
@@ -1,3 +1,17 @@
...
+binary_opts <- c(
+    ":", "\\$", "@", "\\^",
+    "%[^#]*%", "\\+", "-", "\\*", "/",
+    "<", ">", "=", "!", "&", "\\|", "~",
+    "\\?"
+)
+binary_opts_regex <- paste0(binary_opts, collapse = "|")
+binary_opts_ending_regex <- paste0("^[^#]*(", binary_opts_regex, ")\\s*(#.*)?$")
+incomplete_ending_regex <- paste0("^[^#]*(", binary_opts_regex, "|,)\\s*(#.*)?$")
+
```
randy3k, line 14: You may also move to move
https://github.com/REditorSupport/languageserver/blob/7f51365afe5853a48218c30e8d65f5590cef94e9/R/section.R#L176-L180
here

randy3k: Similarly, do you want to define a function for `incomplete_ending_regex`?

renkun-ken: Maybe we could revisit if we find other use cases elsewhere?

PR #655, opened 2024-01-28
Add devcontainer configs

renkun-ken:


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM rocker/r-base
```
eitsupi, line 1: We recommend using `docker.io/library/r-base` (`r-base`) instead of `rocker/r-base`.
`rocker/r-base` is only available on amd64, but `r-base` is a multi-platform image.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+COPY DESCRIPTION .
+
+RUN Rscript -e '                           \
+  install.packages("remotes");             \
+  remotes::install_deps(dependencies = c(  \
+    "Imports",                             \
+    "Config/needs/development"             \
+  ))                                       \
+'
```
eitsupi, line 14: It is recommended to use the `updateContentCommand` instead of doing the build in the Dockerfile;
there is no need to rebuild the image if the DESCRIPTION file is updated.

See rocker-org/devcontainer-features#208

.devcontainer/Dockerfile
renkun-ken: Thanks for the information!

.devcontainer/Dockerfile
renkun-ken: Thanks for the information! Would you like to send a PR for the suggested changes?

eitsupi: Sure, I'll send a PR.

PR #656, opened 2024-01-28
dev(devcontainer): change to install R via rig and using pak

eitsupi:
A follow up for #655

- Change to installing R via rig on a base image with a large amount of tools installed.
- R package installation is done via pak.


renkun-ken (approved):
LGTM Thanks!

PR #756, opened 2026-08-29
Classify function assignment names as function semantic tokens

wmsci:
Hi,

I noticed this issue with the syntax highlighting while using languageserver with the R extension for Visual Studio Code in Cursor. I tried fixing it myself and it works on my machine, so I'm opening this PR in case it is useful.

## Problem

When a function is defined with `fn <- function() {}` and later used as a function call, `fn()`, those two uses of `fn` get different semantic token types.

The function call is a `SYMBOL_FUNCTION_CALL`, so it is marked `function`. The name on the left of the assignment is only a `SYMBOL`, so it is marked `variable`.

Themes colour those types separately, so the same name is coloured differently at the definition and at the function call.

Completion already treats these assigned names as functions (`scope_completion_functs_xpath` in `R/completion.R`).

## Before

SCR-20260829-spjr

## After

SCR-20260829-sqxl

## Fix

After the default parse-data map, reclassify a `SYMBOL` as a `function` token with the `declaration` modifier when it is the left-hand name of:

- `fn <- function() {}`
- `fn = function() {}`
- `fn <- \(x) {}`

The same rule is applied on the XML fallback path.

## Testing

- Added cases in `tests/testthat/test-semantic-tokens.R`

### DIFF
--- a/R/semantic.R
+++ b/R/semantic.R
@@ -69,6 +69,67 @@ empty_semantic_data <- function() {
     )
 }
 
+#' Parse-data ids of symbols assigned `function()` or `\()`
+#'
+#' Same cases as `scope_completion_functs_xpath` in `completion.R`.
+#' @noRd
+function_assignment_symbol_ids <- function(data) {
+    fun_expr_ids <- unique(data$parent[data$token %in% c("FUNCTION", "'\\\\'")])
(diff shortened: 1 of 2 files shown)


renkun-ken (approved):
LGTM Thanks!

PR #759, opened 2026-09-11
protect extract_default_values for namespaced calls

ashiklom:
When an argument default parameter is a namespaced call (`package::function`), `as.character(default_expr[[1]])` is a character vector with multiple elements (`c("::", "package", "function")`. This will cause `func_name == "c"` to error with `condition has length > 1`. However, `identical(func_name, "c")` will safely and correctly handle the fact that `func_name` is not _exactly_ the length-1 character vector "c".

For example:

```r
f <- function(a = stats::mean(stats::rnorm(100))) {
  a
}

default_expr <- formals(f)[["a"]]
# stats::mean(stats::rnorm(100))

is.call(default_expr)
# [1] TRUE

length(default_expr)
# [1] 2

func_name <- as.character(default_expr[[1]])
# [1] "::"    "stats" "mean" 

length(func_name)
# [1] 3

if (func_name == "c") {
  print("yes")
} else {
  print("no")
}
# Error in if (func_name == "c") { : the condition has length > 1

if (identical(func_name, "c")) {
  print("yes")
} else {
  print("no")
}
# [1] "no"
```

I discovered this when encountering a bunch of completion errors in [targets](https://github.com/ropensci/targets). For example, most of the default arguments to `targets::tar_target` are namespaced, which triggers this LSP error when trying to do any kind of completion:

```
> formals(targets::tar_target)
$name

$command

$pattern
NULL

$tidy_eval
targets::tar_option_get("tidy_eval")

$packages
targets::tar_option_get("packages")

$library
targets::tar_option_get("library")

# etc...
```


renkun-ken:
LGTM Thanks!

PR #767, opened 2026-09-15
fix: avoid passing null or zero-length buffers to qsort in reference resolution

randy3k:
### Summary
Fixes undefined behavior reported by CRAN clang-UBSAN and gcc-UBSAN check flavors:
```
navigation.c:144:11: runtime error: null pointer passed as argument 1, which is declared to never be null
/usr/include/stdlib.h:999:30: note: nonnull attribute specified here
    #0 reference_resolve_local_c src/navigation.c:144:5
```

### Changes
- In `src/navigation.c` (`reference_resolve_local_c`):
  - Pre-fill `result` with 0s and return early if `n == 0 || d == 0`, avoiding `R_alloc(0, ...)`.
  - Return early if `n_occurrences == 0 || n_definitions == 0`.
  - Guard `qsort` calls with `if (n_occurrences > 1)` and `if (n_definitions > 1)` to avoid passing null or single-element buffers to `qsort`.
- In `tests/testthat/test-navigation-index.R`:
  - Add test cases covering empty definitions/occurrences and non-positive scope groups.


randy3k:
### R-hub UBSAN Check Results

Verified on R-hub via workflow run [#35002717492](https://github.com/REditorSupport/languageserver/actions/runs/35002717492):

- **`clang-ubsan`**: **`Status: OK`** (`[ FAIL 0 | WARN 0 | SKIP 155 | PASS 4775 ]`)
- **`gcc-asan`** (includes `gcc-UBSAN`): **`Status: OK`** (`[ FAIL 0 | WARN 0 | SKIP 155 | PASS 4775 ]`)

Both checks completed cleanly with 0 errors and 0 warnings.

renkun-ken (approved):
LGTM Thanks!

PR #768, opened 2026-09-15
fix: ensure error and result are mutually exclusive in JSON-RPC responses (#766)

### LINKED ISSUE #766: [BUG REPORT]: Cancellation responses include both `result:null` and `error`, and violate JSON-RPC 2.0
The violation of JSON-RPC 2.0 causes strict clients, like Neovim's built-in LSP, to kill the connection.

**Environment:**
- languageserver 0.3.19
- R 4.6.1 (2026-06-24)
- Linux, x86_64
- Client: Neovim's built-in `vim.lsp`

**Description:**

When a background request is superseded by a newer document version, the server replies with a `ResponseErrorMessage` with JSON content like this:

```json
{"jsonrpc":"2.0","id":2,"result":null,"error":{"code":-32800,"message":"Request superseded by a newer document version"}}
```

This includes **both** `result` (as `null`) and `error`. Per the JSON-RPC 2.0 spec, a Response object must contain exactly one of
`result` or `error`, never both:

> Either the result member or error member MUST be included, but both members MUST NOT be included.

The cause is in the R6 "Response" class here:
https://github.com/REditorSupport/languageserver/blob/b85fcdf8d5853f5ed3d054bc885ea4c9e393a477/R/protocol.R#L97-L107

`result` is unconditionally placed in the `list()` before `error` is conditionally appended. `ResponseErrorMessage$initialize()` never sets `self$result`, so it stays `NULL` — but `list(result = NULL)` in R keeps `result` as an explicit key with a JSON `null` value. `response_to_json()` (`src/json.c: response_json_c`) then serializes that `null` and produces an invalid double-field response.

**Suggested fix:**

In `Response$to_json()` (and/or directly in `ResponseErrorMessage`, via R6 inheritance), only include whichever of `result`/`error` is actually set:

```r
to_json = function() {
    payload <- list(jsonrpc = self$jsonrpc, id = self$id)
    if (!is.null(self$error)) {
        payload$error <- self$error
    } else {
        payload$result <- self$result
    }
    response_to_json(payload)
}
```

Happy to open this as a PR if useful. Thanks

randy3k:
Fixes #766

### Summary
Per JSON-RPC 2.0 (Section 5):
> Either the result member or error member MUST be included, but both members MUST NOT be included.

Previously, `Response()` unconditionally included `result = self` in the initial payload `list()`, so error responses produced both `result: null` and `error: {...}`, causing strict clients like Neovim's built-in `vim.lsp` to terminate the connection.

### Changes
- In `R/protocol.R` (`Response()`):
  - Construct payload with `error` when `!is.null(self)`, and with `result` otherwise.
- In `tests/testthat/test-response-json.R`:
  - Update expected payloads to reflect mutual exclusivity of `result` and `error`.
  - Assert that error responses do not contain `result`, and success responses (including those with `result = NULL`) do not contain `error`.


renkun-ken (approved):
LGTM Thanks!

